PR #6, opened 2022-01-28
changes

lispwarez:


kellerza:
Thanks @lispwarez 

Can you also document the filter you propose? Docs.md

Another option could be to parameterise the filters - use fff[1] or fff[2] if it exists to set interval and samples, then your “now” becomes: `max:2:1`

Why does the HA supervisor not show your PORT with the current tty filter?

lispwarez:
Hey

 

Im an absolute novice with github so not 100% sure how to be doing all this. I have updated the Docs.md and submitted a new pull request.

 

Line 49 of Docs.md seems to be an incomplete sentence. Not 100% sure what you were traying to say.

 

You could add parameters to the filters but I foresee more chance for users to do something wrong and then you will get more ‘bugs’ reported

 

As for the PORT setting I run my home assistant instance in MS Hyper-V which does not have any USB ports exposed to the VM. If you are forcing the settings to match a device but there is no device available you just cant save the settings and then cant start the addon

kellerza:
To add changes you simply push another commit, so what you did seemed to work.

PR #12, opened 2022-02-03
Update issue.md

reedy:


kellerza:
Thx!

i’ve changed the templates to not use the tick boxes, since they show tasks, will merge it with master over the weekend

Issue #10, opened 2022-02-03
USB to RS485 adapter issues?


reedy:
**Issue related to**
- [ ] New Sensors / Sensors not working (existing definitions available [here](https://github.com/kellerza/sunsynk/blob/main/sunsynk/definitions.py))
- [x] Sunsynk Home Assistant Add-On
- [ ] mbusd Home Assistant Add-On
- [ ] Python sunsynk library

**Describe the issue/bug**
Not so much a bug (necessarily), but I guess a support request.

Recently had a Sunsynk 3.6 kWh inverter installed.

I bought one of https://www.ebay.co.uk/itm/393575748700 "RS485-150U" hoping it would work (being premade), but when starting the addon it seems to timeout.

I don't know if it's due to the wrong pinout on the RJ45 end, but I did try `/dev/ttyACM0`, `/dev/ttyAMA0` and `/dev/serial1`; all of which failed similarly.

I have plugged it into the BMS485 (top left) port; it seems for a lot of things the 5.5 and 3.6 are similar.

Got a different adapter coming tomorrow from Amazon - https://smile.amazon.co.uk/dp/B06Y1JTGZX which will make it easier to make sure the pinout is right, so this request might be invalid at that point...

Trying to avoid having to shell out for one of https://solar-assistant.io/shop/products/sunsynk_rs485

Happy to try a few other things (and/or doing this outside Home Assistant.. On a vanilla Ubunutu/Raspberry Pi OS/Debian rpi install), such as the library as a standalone if it helps.

Thanks!

**Expected behavior**
Not crash on startup? :)

**Home Assistant version (if applicable)**
2022.2.0

**Additional context**

**Logs**

```
[s6-init] making user provided files available at /var/run/s6/etc...exited 0.
[s6-init] ensuring user provided files have correct perms...exited 0.
[fix-attrs.d] applying ownership & permissions fixes...
[fix-attrs.d] done.
[cont-init.d] executing container initialization scripts...
[cont-init.d] done.
[services.d] starting services
[services.d] done.
2022-02-03 18:20:43,458 INFO    Loading HASS OS configuration
2022-02-03 18:20:43,460 INFO    Filter *last used for ['total_pv_power', 'total_battery_charge', 'total_battery_discharge', 'total_grid_export', 'total_grid_import']
2022-02-03 18:20:43,460 INFO    Filter *step used for ['battery_soc']
2022-02-03 18:20:43,469 INFO    Protocol made connection.
[...]

PR #14, opened 2022-02-04
Add and use some images

reedy:


kellerza:
Great, this should make it easier for new users!

can you add your inverter model to the table as well?

reedy:
> can you add your inverter model to the table as well?

I can, yeah, but I didn't want to mark it as working as per my comments in https://github.com/kellerza/sunsynk/issues/10#issuecomment-1030200209 it seems to be connecting, but unable to get any data out of it just yet...

kellerza (approved):

PR #22, opened 2022-02-23
Update DOCS.md

reedy:


kellerza:
Thanks @reedy, I made some significant updates to the docs, so this one is stale.

The new docs does have a bit more punctuation :-)

PR #23, opened 2022-03-04
Update README.md

JohanViljoen:
Got it mostly working on my Deye 8kW inverter. Still lots of timeouts, but at least something gets through occasionally.


kellerza (commented):
Thanks!

PR #35, opened 2022-04-20
Added sensor

agtconf:
Hi.

I have found and used this sensor in Jlopez's implementation to interrogate the inverter through the wifi connection. It is uploaded to github and the file with the specification is this

https://github.com/jlopez77/DeyeInverter/blob/main/DYRealTime.txt

the values that the sensor can have are 1 and 0

{
        "titleZH": "并网状态",
        "titleEN": "Grid-connected Status",
        "registers": ["0x00C2"],
        "interactionType": 2,
        "parserRule": 1,
        "optionRanges": [
          {
            "key": 0,
            "value": "Off-Grid"
          },
          {
            "key": 1,
            "value": "On-Grid"
          }
        ],
        "ratio": 1,
        "unit": ""
      },

I don't know if this is the best type of sensor.

It works in turbo-energy and deye inverters.
I don't know python but I think the syntax is something like this.

Regards.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ There are several inverters that are rebranded Deye inverters, so you might have
 | Sunsynk 5.5kW  | Hubble AM-2       | beta/all | @kellerza     | BMS485 (top left)          |
 | Sunsynk 8.8kW  | BSL 8.2 kWH       | 0.0.8    | @dirkackerman | RS485 (1 in image below)   |
 | Deye  8kW      | Pylontech US3000C | 0.1.3dev | @Kladrie      | RS485 (top left)           |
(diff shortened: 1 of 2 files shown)


kellerza:
Thanks! I'll merge and you can try the sensor. To date, I've used `grid_frequency` to check if the grid is connected

agtconf:
Thank you!
I'm not able to add sensor to HA.
I upgraded to sunsynk 0.1.5 via pip3 with the following command

sudo pip3 install sunsynk --upgrade 

But whe I add 

- grid_connected_status

to the HA-sunsynk configuratión, it doesn't appear as a sensor when I try to add it.

Regards.

kellerza:
Only the -dev version uses 0.1.5 at the moment

Issue #33, opened 2022-04-11
Sunsynk data not reading


marvil1961:
I have HA installed on a Ubuntu server inside virtualbox as per the installation instructions ha installation on Linux (https://www.home-assistant.io/installation/linux/#install-home-assistant-operating-system)

Have battled to get the usb to RS485 converter to be seen by HA but found that the usb device must first be enabled in Virtualbox and were then quite exited that HA will work with my 8kW Sunsynk Inverter but it was not meant to be - communication is still not happening. Cable connection I think is correct Oninverter Pin 1&2 to B&A on Usb converter

on the Modbus TCP to Modbus RTU Gateway Add-on my Log info is as follows:
DEVICE: /dev/serial/by-id/usb-1a86_USB2.0-Ser_-if00-port0

[19:21:07] INFO: Starting mbusd -d -L - -v 3 -p /dev/ttyUSB0 -s 9600 -m 8N1 -P 502
11 Apr 2022 19:21:07 mbusd-0.5.1 started...
11 Apr 2022 19:21:07 tty: trying to open /dev/ttyUSB0 (speed 9600 mode 8N1)
11 Apr 2022 19:21:59 tty: error in read() (port closed)
11 Apr 2022 19:21:59 trying to re-open tty...
11 Apr 2022 19:21:59 tty re-opening...
11 Apr 2022 19:21:59 tty: trying to open /dev/ttyUSB0 (speed 9600 mode 8N1)
11 Apr 2022 19:21:59 tty re-opened.

The above looks like this part is working

on the Sunsynk Inverter Add-on the log file looks like this and I get Timeout Readings 

[s6-init] making user provided files available at /var/run/s6/etc...exited 0.
[s6-init] ensuring user provided files have correct perms...exited 0.
[fix-attrs.d] applying ownership & permissions fixes...
[fix-attrs.d] done.
[cont-init.d] executing container initialization scripts...
[cont-init.d] done.
[services.d] starting services
[services.d] done.
2022-04-11 19:26:04,237 INFO    Loading HASS OS configuration
2022-04-11 19:26:04,294 INFO    Filter *last used for battery_soc, total_battery_charge, total_battery_discharge, total_grid_export, total_grid_import, total_pv_power
2022-04-11 19:26:04,305 INFO    Protocol made connection.
2022-04-11 19:26:04,306 INFO    Connected to /dev/ttyUSB0
2022-04-11 19:26:04,364 INFO    ############################################################
[...]

PR #43, opened 2022-07-12
Add RS485 line voltage troubleshooting tips

mikee47:
See https://github.com/kellerza/sunsynk/issues/33#issuecomment-1177198808


kellerza:
Thanks @mikee47 !

Issue #37, opened 2022-05-29
Enhancement: Writable sensors


Ivan-L:
### Enhancement related to

Sunsynk Home Assistant Add-On

### Description

Hi @kellerza, I see you recently introduced the concept of a RWSensor so I can see that you are planning to support this in the future. Have you got a branch or any in-progress work for this? If not, I'd like to take a stab at it to introduce a missing sensor and make it writable based on a [Select Entity](https://developers.home-assistant.io/docs/core/entity/select/). The sensor is number 243 which allows you to toggle between Battery or Load priority for PV production.

Happy to hear your thoughts!

kellerza:
Writing registers are supported by profiles today. So if you create a new profile and add your sensor to the array it should work.

Actual values being written is stored in yaml files under /share/addon_name/. Select UPDATE in the frontend and the current value (or multiple values in case of many sensors) will be written

See [here](https://github.com/kellerza/sunsynk/blob/v0.1.5/hass-addon-sunsynk-dev/profiles.py#L224)

Ivan-L:
Yeah I have been using profiles for a while, they are extremely useful!
<img width="651" alt="image" src="https://user-images.githubusercontent.com/16413523/171053423-6868eae0-f516-42de-a107-91edc5db0328.png">

What I was thinking was that having certain entities editable can be complementary to profiles and actually make creating them easier from Home Assistant. For example, whilst on vacation (therefore no access to the inverter), any time I want to modify the battery percentage of one of the programs (Prog1 Capacity-Prog6 Capacity) I need to access the /share/addon_name, copy and paste one of the profile files, edit the values in it, rename it and then reload. If however I could edit them via the Home Assistant UI, that would be so much easier. Same for the above sensor 243 - currently a profile would need to exist already for the register to be writable.

Here is where the inspiration came from, having device config entities for some/all writable inverter registers:
https://developers.home-assistant.io/blog/2021/10/26/config-entity/#entity-categories

kellerza:
Ok, that is an interesting idea. `entity_category` seems to be part of the MQTT autodiscovery options, so it should be possible.

If you decide to add all RWSesnors / descendants they will also have to be read as well, and for that the round robin filter should be ok

Ivan-L:
I'll do a proof of concept first to see if the idea is viable and report here.

Ivan-L:
[...]

PR #48, opened 2022-07-19
Introduce NumberRWSensor

Ivan-L:
This is the first of a series of PRs addressing #37.

This first PR allows for certain sensors to be changed to be of type `NumberRWSensor` which will make them editable in Home Assistant. Furthermore, the `min` and `max` values can be static (with a default of 0 and 100 respectively) or they can depend on another sensor. What still needs to be done is that if a `NumberRWSensor` depends on other sensors for its `min` or `max` values and those other sensors change, the MQTT Home Assistant number entity needs to be updated with the new `min` or `max` values. This would be tackled as part of a subsequent PR.

Will add tests shortly to increase code coverage.


hass-addon-sunsynk-dev/run.py
```diff
@@ -208,6 +240,29 @@ async def read_sensors(
...
+    _LOGGER.info(
+        "Writing sensor %s: %s=%s  [old %s]",
+        sensor.name,
+        sensor.id,
+        newv,
+        sensor.reg_value,
+    )
+    sensor.reg_value = newv
+    await SUNSYNK.write_sensor(sensor)
+    await read_sensors([sensor])
```
Ivan-L, line 260: Could this happen at the exact time as `poll_sensors()` and cause problems if we're trying to do two reads at the same time?

hass-addon-sunsynk-dev/run.py
```diff
@@ -49,27 +49,58 @@ async def publish_sensors(sensors: List[Filter], *, force: bool = False) -> None
...
+                    command_topic=f"{SS_TOPIC}/{OPT.sunsynk_id}/{sensor.id}_set",
+                    min=float(await static_or_sensor_value(sensor.min)),
+                    max=float(await static_or_sensor_value(sensor.max)),
+                    unit_of_measurement=sensor.unit,
+                    unique_id=f"{OPT.sunsynk_id}_{sensor.id}",
+                    device=dev,
+                    on_change=create_on_change_handler(filt, int),
+                )
+            )
+        else:
```
kellerza, line 94: Please use a continue here instead of the else. So by default it simply falls through

hass-addon-sunsynk-dev/run.py
```diff
@@ -165,6 +203,21 @@ def startup() -> None:
...
+
+    # Add any sensors that are depended upon which were not specified in OPT.sensors
+    for name, sen in sens_dependencies.items():
+        if name not in sens and sen != RATED_POWER:  # Rated power does not change
+            fstr = suggested_filter(sen)
+            msg[f"*{fstr}"].append(name)  # type: ignore
+            SENSORS.append(getfilter(fstr, sensor=sen))
+
+            _LOGGER.info("Added sensor %s as other sensors depend on it", name)
+
```
kellerza, line 220: Can't we rather define the min&max in the configuration if its different than the default in the definition?

Is there any case where the min/max could actually change?

hass-addon-sunsynk-dev/run.py
kellerza: Not with async code, you only have a single thread here. Read is only scheduled for the next second

The read might affect the filter (the filter makes the assumption a read only happens on the second). But these writable entities should all use the round-robin filter so its not a big issue

sunsynk/definitions.py
```diff
@@ -126,6 +127,8 @@
 ##########
 # General
 ##########
+RATED_POWER = Sensor((16, 17), "Rated power", WATT, 0.1)
```
kellerza, line 130: You now have 2xconstants name `RATED_POWER` in different files...

kellerza (changes requested):

hass-addon-sunsynk-dev/mqtt.py
```diff
@@ -128,6 +128,26 @@ class SelectEntity(Entity):
...
+
+    on_change: Callable = attr.field(default=None)
+
+    _path = "number"
+
+    def __attrs_post_init__(self) -> None:
+        """Init the class."""
+        # Override device_class to None
+        # https://github.com/home-assistant/core/issues/74533
+        self.device_class = None
```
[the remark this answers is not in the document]
kellerza: Post init can be used to update the inherited class, so probably ok

hass-addon-sunsynk-dev/run.py
kellerza: Scratch the previous comment. Looking at the code, the main issue will be that you cannot guarantee BATTERY_LOW_CAPACITY will be read **before** you do sensor discovery. 

Once you've done sensor discovery changing the battery low capacity will not update the discovered elements (I'm not sure we want to do this).

Maybe we can add RATED_POWER and BATTERY_LOW CAPACITY to be read on startup, so define a STARTUP_SENSORS.

STARTUP_SENSORS = (
SERIAL, RATED_CAPACITY, BATTERY_LOW_CAPACITY
)

Ivan-L:
> Thanks @Ivan-L probably only the question on the startup sensors remaining

PR updated with `STARTUP_SENSORS` added and dependent sensors no longer being added to `SENSORS` collection

kellerza (approved):

kellerza:
Made a minor change to building the STARTUP_SENSORS, but his looks good!

hass-addon-sunsynk-dev/run.py
```diff
@@ -49,18 +51,47 @@ async def publish_sensors(sensors: List[Filter], *, force: bool = False) -> None
...
         identifiers=[OPT.sunsynk_id],
         name=f"Sunsynk Inverter {serial}",
-        model=f"Inverter {serial}",
+        model=f"{int(rated_power/1000)}kW Inverter {serial}",
         manufacturer="Sunsynk",
     )
 
+    def create_on_change_handler(filt: Filter, value_func: Callable):
+        async def _handler(value):
+            await write_sensor(filt, value_func(value))
```
kellerza, line 66: Maybe you can add some async debouncing here. 
With the profiles it only writes in the main loop and these are at least 1 second apart

Ivan-L:
Hi @kellerza, I have updated the PR to solve for the timeouts as can be seen in the comments. I cannot think of anything more to add to this PR. If you are happy with the changes and there are no outstanding issues, you are welcome to merge it in.

kellerza:
Thanks @Ivan-L ! 🎉

Ivan-L:
Sorry @kellerza I submitted an old review so my old comments came through 😊

PR #52, opened 2022-07-31
Added SelectRWSensor and priority_mode

Ivan-L:
This is the second PR addressing #37.

A `SelectRWSensor` has been introduced which supports an `options` dictionary of register value to display value pairs. This sensor is then used by the dev add-on to create a Home Assistant [MQTT Select](https://www.home-assistant.io/integrations/select.mqtt/).

`RWSensor` also gained come common writable sensor functionality through the `update_reg_value` method which has the common steps to update the register value of the sensor. Subclasses merely need to map an incoming `value` to a register value. Any of these methods can be renamed if needed.

`NumberRWSensor` was also enhanced to deal with `factor` and to validate the incoming `value` against `min_value` and `max_value` when creating a register value.

### DIFF
--- a/hass-addon-sunsynk-dev/run.py
+++ b/hass-addon-sunsynk-dev/run.py
@@ -71,23 +71,25 @@ async def hass_discover_sensors(serial: str, rated_power: float) -> None:
     for filt in SENSORS:
         sensor = filt.sensor
 
-        entity_name = f"{OPT.sensor_prefix} {sensor.name}".strip()
         state_topic = f"{SS_TOPIC}/{OPT.sunsynk_id}/{sensor.id}"
-        command_topic = f"{SS_TOPIC}/{OPT.sunsynk_id}/{sensor.id}_set"
-        unique_id = f"{OPT.sunsynk_id}_{sensor.id}"
+        command_topic = f"{state_topic}_set"
+
+        ent = {
+            "device": dev,
+            "entity_category": "config" if isinstance(sensor, RWSensor) else "",
+            "name": f"{OPT.sensor_prefix} {sensor.name}".strip(),
+            "state_topic": state_topic,
+            "unique_id": f"{OPT.sunsynk_id}_{sensor.id}",
+            "unit_of_measurement": sensor.unit,
+        }
 
         if isinstance(sensor, NumberRWSensor):
             ents.append(
                 NumberEntity(
-                    name=entity_name,
-                    entity_category="config",
-                    state_topic=state_topic,
+                    **ent,
                     command_topic=command_topic,
                     min=float(sensor.min_value),
                     max=float(sensor.max_value),
-                    unit_of_measurement=sensor.unit,
-                    unique_id=unique_id,
-                    device=dev,
                     on_change=create_on_change_handler(filt, int),
                 )
             )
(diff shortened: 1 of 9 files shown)


sunsynk/sensor.py
```diff
@@ -197,6 +249,10 @@ def update_value(self) -> None:
         sval = str(self.reg_value[0])
         self.value = f"{sval[:-2]}:{sval[-2:]}"
 
+    def value_to_reg(self, value: Any) -> int | Tuple[int, ...]:
+        """Get the reg value from a display value."""
+        raise NotImplementedError()
```
Ivan-L, line 254: This has already been implemented in another branch as part of some larger `TimeRWSensor` enhancements. This exception will not be hit because `TimeRWSensor` is not writable for now in run.py.

hass-addon-sunsynk-dev/run.py
```diff
@@ -71,29 +71,49 @@ def _handler(value):
     for filt in SENSORS:
         sensor = filt.sensor
 
+        entity_name = f"{OPT.sensor_prefix} {sensor.name}".strip()
```
kellerza, line 74: There seems to be lots of duplication here.
Can you extract all the common attributes? 

```python
ent = {
entity_name: f"{OPT.sensor_prefix} {sensor.name}".strip(),
device=dev
}
```

Then add specific values in the various if statements and call it with the options, i.e.:

```python
NumberEntity(**ent)
NumberEntity(**ent)
SelectRWSensor(**ent)
```

You can even delete keys from options if not required for that sensor (preferably dont use this too often)

Ivan-L: Excellent, I wondered if there was a better way to do this and fully agree. Python is not my first language as you can tell :) Will update tonight.

Ivan-L: It works but one has to quote the dictionary keys to avoid a `Pylint(E0602:undefined-variable)` error.

image
image

Here is the full version with quoted dictionary keys. All duplication has been removed, which is great. The only minor issue is that fields in `ent` are quoted whilst entity-specific fields are not since they are passed in as named arguments, but it seems to make no difference with validation - having an invalid dictionary key in `ent` or an invalid named argument when constructing the specific entity yields the same TypeError at runtime.

```python
ent = {
    "device": dev,
    "name": f"{OPT.sensor_prefix} {sensor.name}".strip(),
    "state_topic": f"{SS_TOPIC}/{OPT.sunsynk_id}/{sensor.id}",
    "unique_id": f"{OPT.sunsynk_id}_{sensor.id}",
    "unit_of_measurement": sensor.unit,
}
if isinstance(sensor, RWSensor):
    ent = {
        **ent,
        "command_topic": f"{SS_TOPIC}/{OPT.sunsynk_id}/{sensor.id}_set",
        "entity_category": "config",
    }

if isinstance(sensor, NumberRWSensor):
    ents.append(
        NumberEntity(
            **ent,
            min=float(sensor.min_value),
            max=float(sensor.max_value),
            on_change=create_on_change_handler(filt, int),
        )
    )
    continue

if isinstance(sensor, SelectRWSensor):
    ents.append(
        SelectEntity(
            **ent,
            options=sensor.available_values(),
            on_change=create_on_change_handler(filt, str),
        )
    )
    continue

ents.append(SensorEntity(**ent))
```

PR #53, opened 2022-08-05
Made TimeRWSensor writable

Ivan-L:
This PR enhances `TimeRWSensor` to be writable as part of addressing #37. The Home Assistant [MQTT Select](https://www.home-assistant.io/integrations/select.mqtt/) integration is used as the UI component since [Input Datetime](https://www.home-assistant.io/integrations/input_datetime/) is not available via MQTT.

The sensor also now returns a list of available values based on the configured `min` and `max` sensors. The available values do wrap across midnight. The step value is currently 15 minutes and can be made into an add-on option if desirable.

The affected sensors are the Prog1..6 Time sensors and the HA entity type changes from `sensor.x` to `select.x`.

**Important:** Updating the values of the Prog1..6 Time sensors though the UI will _not_ cause the `min` or `max` values for the dependent sensors to get updated. It now becomes crucial for sensor updates to cause HASS discovery info to be updated. A subsequent PR will add this functionality, or the functionality can be added to this PR.

image


sunsynk/definitions.py
```diff
@@ -165,13 +165,26 @@
...
+PROG3_TIME = TimeRWSensor(252, "Prog3 Time")
+PROG4_TIME = TimeRWSensor(253, "Prog4 Time")
+PROG5_TIME = TimeRWSensor(254, "Prog5 Time")
+PROG6_TIME = TimeRWSensor(255, "Prog6 Time")
+PROG1_TIME.min_max(PROG6_TIME, PROG2_TIME)
+PROG2_TIME.min_max(PROG1_TIME, PROG3_TIME)
+PROG3_TIME.min_max(PROG2_TIME, PROG4_TIME)
+PROG4_TIME.min_max(PROG3_TIME, PROG5_TIME)
+PROG5_TIME.min_max(PROG4_TIME, PROG6_TIME)
+PROG6_TIME.min_max(PROG5_TIME, PROG1_TIME)
```
Ivan-L, line 179: I added a convenient method to be able to set the min and max sensors in one go. If there is a more elegant way to set up the dependencies taking into account that variables cannot be used before they are declared, then I'd happily implement any suggestion.

sunsynk/sensor.py
```diff
@@ -241,17 +241,72 @@ def update_value(self) -> None:
...
+        if val % step != 0:
+            yield val
+        stop = end if start <= end else end + modulo
+        for i in range(start, stop, step):
+            yield i % modulo
+        if start == end or start != end % modulo:
+            yield end
+
+    @staticmethod
+    def __reg_to_timespan(reg_value: int) -> tuple[int, int]:
```
kellerza, line 304: Can we keep these methods with a single underscore?
Double have some special use cases on inheritance, but here your intention seems only that this is a private method, correct?

sunsynk/sensor.py
```diff
@@ -241,17 +241,72 @@ def update_value(self) -> None:
...
+            return hours * 60 + mins
+
+        min_val = reg_to_minutes(self.min.reg_value[0]) if self.min else 0
+        max_val = reg_to_minutes(self.max.reg_value[0]) if self.max else full_day
+        val = reg_to_minutes(self.reg_value[0])
+
+        time_range = self.__time_range(min_val, max_val, val, step_minutes, full_day)
+
+        return list(
+            map(lambda minutes: self.__format(*divmod(minutes, 60)), time_range)
```
kellerza, line 266: What about a small helper class for time to hide all these conversion complexities? Its also easier to unit test that way

You can store the value as (total)minutes (even mins,hrs if you wish) and have a property for regval and one for str?

```python
class ssTime:
  minutes: int
   
  @property
  def regval(self):
      return ..self.minutes...
       
  @regval.setter
  def regval(self, rval):
      self.minutes = ...rval...
```

kellerza (commented):
Thanks @Ivan-L some minor comments

sunsynk/sensor.py
kellerza: The helper class might not be the correct approach (since the .value / .reg_value is used by inherited methods), but you can potentially still use property get/setter to do the conversions

hass-addon-sunsynk-dev/CHANGELOG.md
```diff
@@ -1,4 +1,14 @@
 # Changelog
+## **2022.08.05-0.2.3** - 2022-08-05
```
kellerza:
```suggestion
## **2022.08.19-0.2.3** - 2022-08-05
```

hass-addon-sunsynk-dev/mqtt.py
```diff
@@ -63,6 +63,7 @@ class Entity:
     expire_after: int = attr.field(default=0)  # unavailable if not updated
     enabled_by_default: bool = attr.field(default=True)
     entity_category: str = attr.field(default="")
+    icon: str = attr.field(default="")
```
kellerza, line 66: Is the idea to send it in discovery info?

sunsynk/sensor.py
```diff
@@ -241,17 +241,90 @@ def update_value(self) -> None:
             _LOGGER.error("Could not decode temperature: %s", err)
 
 
+class Minutes(int):
+    """Deals with inverter time format conversion complexities."""
+
+    @classmethod
+    def from_str_value(cls, str_value: str) -> Minutes:
+        """Create from a string in hh:mm format."""
+        return cls(int(str_value[:-3]) * 60 + int(str_value[-2:]))
```
kellerza, line 250: A suggestion for a property setter below

```suggestion
    @str_value.setter
    def str_value(self, value: str):
        """Parse a string in hh:mm format."""
        (hour, _, min) = value.partition(":") 
        self.minutes = int(hour) * 60 + int(min)
```

Inheriting from a base value like `int` is not ideal, since the value will get replace on each assignment/math operation. So best to store the value of `minutes` in the class, thne it becomes `self.minutes`

```python
class Minutes:
    minutes: int
```

sunsynk/sensor.py
```diff
@@ -241,17 +241,90 @@ def update_value(self) -> None:
...
+
+    @classmethod
+    def from_reg_value(cls, reg_value: int) -> Minutes:
```
kellerza, line 253: The reg_value can also be rework so that you can assign to it with `@reg_value.setter`

sunsynk/sensor.py
```diff
@@ -241,17 +241,90 @@ def update_value(self) -> None:
...
+        if isinstance(self.min, TimeRWSensor):
+            sensors.append(self.min)
+        if isinstance(self.max, TimeRWSensor):
+            sensors.append(self.max)
+        return sensors
+
+    def min_max(self, min_value: TimeRWSensor, max_value: TimeRWSensor) -> None:
+        """Set the min and max values."""
+        self.min = min_value
+        self.max = max_value
```
kellerza, line 307: taking the other suggestion into account, we probably don't need this. It is also only used once

kellerza (changes requested):

sunsynk/sensor.py
kellerza: :+1: Lets stay with the helper, since you've implemented it

hass-addon-sunsynk-dev/mqtt.py
kellerza: Lets remove this as its not being used

sunsynk/definitions.py
```diff
@@ -165,13 +165,27 @@
 _SENSORS.append(
     SelectRWSensor(243, "Priority Mode", options={0: "Battery first", 1: "Load first"})
 )
+PROG1_TIME = TimeRWSensor(250, "Prog1 Time")
+PROG2_TIME = TimeRWSensor(251, "Prog2 Time", min=PROG1_TIME)
+PROG3_TIME = TimeRWSensor(252, "Prog3 Time", min=PROG2_TIME)
+PROG4_TIME = TimeRWSensor(253, "Prog4 Time", min=PROG3_TIME)
+PROG5_TIME = TimeRWSensor(254, "Prog5 Time", min=PROG4_TIME)
+PROG6_TIME = TimeRWSensor(255, "Prog6 Time", min=PROG5_TIME)
+PROG1_TIME.min = PROG6_TIME
```
kellerza, line 174: Are you sure prog1 should have a minimum of prog6? - have asked this question previously, without answer.

sunsynk/sensor.py
kellerza: Please check this

kellerza (changes requested):

Ivan-L:
Hi @kellerza based on the last review I think we missed each other, so I'm just typing this up here to ensure that no comments are missed.

1. The `icon` field that was added to the MQTT `Entity` class - this was actually being used when mapping `TimeRWSensor` to `SelectEntity`. I did previously comment on this with screenshots, but maybe you missed it? Nevertheless I have removed the icon support.
2. Regarding the min for Prog 1 - I did comment here too:
https://github.com/kellerza/sunsynk/pull/53#discussion_r947506917
https://github.com/kellerza/sunsynk/pull/53#discussion_r947866930
https://github.com/kellerza/sunsynk/pull/53#discussion_r950094957
In my view the min for Prog 1 should not be 0:00. My inverter UI does allow me to decrease prog 1 to be before midnight as per above comments. To me having a dynamic min of Prog 6 for Prog 1 does make sense. Can you check on your inverter whether you can decrease Prog 1 to be before midnight (assuming your Prog 6 is before midnight too)?
3. I have reworked the time class helper with the suggestions. Apologies I initially misunderstood you. Please have a look.

kellerza:
Even with your direct links, I cannot see that you have ever responded to any question I've asked ?!

Ivan-L:
My PR is not in draft mode (unlike my other one):
image

I usually responded to your questions almost immediately. Maybe I did something incredibly dumb. My comments to appear as Pending. For example:
image

Ivan-L:
Well this was a facepalm moment for me. You should now be able to see all of my comments. Sorry about that.

kellerza:
Now it starts making some sense 😂

sunsynk/definitions.py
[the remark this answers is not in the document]
kellerza: Interesting, let me check this tomorrow!

sunsynk/sensor.py
kellerza: PAscal case would just be `SsTime` - starts with caps

hass-addon-sunsynk-dev/mqtt.py
kellerza: Ok, but not sure it was actually sent via MQTT though?
But lets add this in a new PR once we're done here

kellerza:
All these comments are very confusing now, but let me just double check that my inverter can wrap and then we can merge (and maybe even start a fresh PR for the options)

hass-addon-sunsynk-dev/mqtt.py
Ivan-L: Yep it was sent via MQTT. I can definitely put up another PR which just deals with icons.

kellerza:
My inverter allows any input and there are no real min/max time values enforced here. If this works is another question, but guess we are already doing a bit better (or enforcing a bit more consistency) here

PR #54, opened 2022-08-07
Update HASS auto discovery info when sensor values change & add RWSensor icons

Ivan-L:
This PR builds on the work done in #37 to allow for updating of HASS auto discovery info if a sensor, upon which other sensors depend for their min/max values, changes.

The changes here solve for two issues:

1. If the user of the add-on has added a sensor which depends on another sensor for the min/max values but the user has not added the latter sensor, it automatically gets added as a hidden sensor so that its value can be read in the main loop and the former sensor's HASS discovery info can be updated.
2. If the user walks over to the inverter and changes the sensor value, or the value is changed via the Home Assistant UI (if the sensor is a subclass of `RWSensor`), the change is picked up by the add-on and the HASS discovery info is updated.

Furthermore, the writable battery capacity sensors are now instances of RWSensor to distinguish them from the battery SOC sensors to fix an issue where the Home Assistant would pick the first (wrong) sensor with a device class of `battery` as the device's SOC.

Finally, some icons for RWSensor entities were added.

### DIFF
--- a/hass-addon-sunsynk-dev/mqtt.py
+++ b/hass-addon-sunsynk-dev/mqtt.py
@@ -63,14 +63,12 @@ class Entity:
     expire_after: int = attr.field(default=0)  # unavailable if not updated
     enabled_by_default: bool = attr.field(default=True)
     entity_category: str = attr.field(default="")
+    icon: str = attr.field(default="")
 
     _path = ""
 
     def __attrs_post_init__(self) -> None:
         """Init the class."""
-        if not self.device_class:
-            # Try device_class from unit_of_measurement
-            self.device_class = hass_device_class(unit=self.unit_of_measurement)
         if not self.state_class and self.device_class == "energy":
             self.state_class = "total_increasing"
 
@@ -140,12 +138,6 @@ class NumberEntity(Entity):
     on_change: Callable = attr.field(default=None)
 
     _path = "number"
-
-    def __attrs_post_init__(self) -> None:
-        """Init the class."""
-        # Override device_class to None
-        # https://github.com/home-assistant/core/issues/74533
-        self.device_class = None
 
 
 class MQTTClient:
@@ -337,6 +329,16 @@ def hass_device_class(*, unit: str) -> str:
         "A": "current",
         "°C": "temperature",
         "%": "battery",
+    }.get(unit, "")
+
+
+def hass_default_rw_icon(*, unit: str) -> str:
+    """Get the HASS default icon from the unit."""
+    return {
+        "W": "mdi:flash",
+        "V": "mdi:sine-wave",
+        "A": "mdi:current-ac",
+        "%": "mdi:battery-lock",
     }.get(unit, "")
 
 
--- a/hass-addon-sunsynk-dev/run.py
+++ b/hass-addon-sunsynk-dev/run.py
@@ -70,7 +70,6 @@ async def hass_discover_sensors(serial: str, rated_power: float) -> None:
         model=f"{int(rated_power/1000)}kW Inverter {serial}",
         manufacturer="Sunsynk",
     )
-    # pylint: disable=import-outside-toplevel
     global DEVICE  # pylint: disable=global-statement
     DEVICE = dev
 
@@ -134,14 +133,10 @@ def create_entities(sensors: list[Filter], dev: Device) -> list[Entity]:
         }
 
         if isinstance(sensor, RWSensor):
-            ent.update(
-                {
-                    "entity_category": "config",
-                    "icon": hass_default_rw_icon(unit=sensor.unit),
-                }
-            )
+            ent["entity_category"] = "config"
+            ent["icon"] = hass_default_rw_icon(unit=sensor.unit)
         else:
-            ent.update({"device_class": hass_device_class(unit=sensor.unit)})
+            ent["device_class"] = hass_device_class(unit=sensor.unit)
 
         if isinstance(sensor, NumberRWSensor):
             ents.append(
@@ -167,7 +162,7 @@ def create_entities(sensors: list[Filter], dev: Device) -> list[Entity]:
             continue
 
         if isinstance(sensor, TimeRWSensor):
-            ent.update({"icon": "mdi:clock"})
+            ent["icon"] = "mdi:clock"
             ents.append(
                 SelectEntity(
                     **ent,
--- a/hass-addon-sunsynk-dev/run.py
+++ b/hass-addon-sunsynk-dev/run.py
@@ -23,12 +23,12 @@ from sunsynk.sunsynk import Sensor, Sunsynk
 _LOGGER = logging.getLogger(__name__)
 
 
+DEVICE: Device = None
+HASS_DISCOVERY_INFO_UPDATE_QUEUE: Dict[str, Filter] = {}
 SENSORS: List[Filter] = []
-
+SENSOR_WRITE_QUEUE: Dict[str, Tuple[Filter, Any]] = {}
 SERIAL = ALL_SENSORS["serial"]
 STARTUP_SENSORS: List[Filter] = []
-SENSOR_WRITE_QUEUE: Dict[str, Tuple[Filter, Any]] = {}
-
 SUNSYNK: Sunsynk = None  # type: ignore
 
 
@@ -54,13 +54,39 @@ async def publish_sensors(sensors: List[Filter], *, force: bool = False) -> None
 
 async def hass_discover_sensors(serial: str, rated_power: float) -> None:
     """Discover all sensors."""
-    ents: List[Entity] = []
     dev = Device(
         identifiers=[OPT.sunsynk_id],
         name=f"Sunsynk Inverter {serial}",
         model=f"{int(rated_power/1000)}kW Inverter {serial}",
         manufacturer="Sunsynk",
     )
+    # pylint: disable=import-outside-toplevel
+    global DEVICE  # pylint: disable=global-statement
+    DEVICE = dev
+
+    ents = create_entities(SENSORS, dev)
+
+    profile_add_entities(entities=ents, device=dev)
+
+    await MQTT.connect(OPT)
+    await MQTT.publish_discovery_info(entities=ents)
+
+
+async def hass_update_discovery_info() -> None:
+    """Update discovery info for existing sensors"""
+    if not HASS_DISCOVERY_INFO_UPDATE_QUEUE:
+        return
+
+    ents = create_entities(HASS_DISCOVERY_INFO_UPDATE_QUEUE.values(), DEVICE)
+    HASS_DISCOVERY_INFO_UPDATE_QUEUE.clear()
+
+    await MQTT.connect(OPT)
+    await MQTT.publish_discovery_info(entities=ents, remove_entities=False)
+
+
+def create_entities(sensors: list[Filter], dev: Device) -> list[Entity]:
+    """Create HASS entities out of an existing list of filters"""
+    ents: List[Entity] = []
 
     def create_on_change_handler(filt: Filter, value_func: Callable):
         def _handler(value):
@@ -68,7 +94,7 @@ async def hass_discover_sensors(serial: str, rated_power: float) -> None:
 
         return _handler
 
-    for filt in SENSORS:
+    for filt in sensors:
         sensor = filt.sensor
 
         state_topic = f"{SS_TOPIC}/{OPT.sunsynk_id}/{sensor.id}"
@@ -119,10 +145,7 @@ async def hass_discover_sensors(serial: str, rated_power: float) -> None:
 
         ents.append(SensorEntity(**ent))
 
-    profile_add_entities(entities=ents, device=dev)
-
-    await MQTT.connect(OPT)
-    await MQTT.publish_discovery_info(entities=ents)
+    return ents
 
 
 def setup_driver() -> None:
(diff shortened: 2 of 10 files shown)


Ivan-L:
@kellerza I cannot seem to add you as a reviewer, but please would you check out this PR?

hass-addon-sunsynk-dev/run.py
```diff
@@ -203,18 +239,36 @@ def setup_sensors() -> None:
...
 
-        SENSORS.append(getfilter(fstr, sensor=sen))
+        filt = getfilter(fstr, sensor=sen)
+        SENSORS.append(filt)
+        filters[sen.id] = filt
 
         if isinstance(sen, NumberRWSensor):
-            startup_sens.update(d.id for d in sen.dependencies())
+            for dep in sen.dependencies():
+                sens_dependencies[dep.id].append(sen)
```
Ivan-L, line 248: Note that we're constructing a reverse map of dependencies here.
So if sensor A depends on sensor S and sensor B depends on sensor S, then we will have S with two dependants - A and B.

sunsynk/sensor.py
```diff
@@ -42,11 +42,12 @@ class Sensor:
     name: str = attr.field()
     unit: str = attr.field(default="")
     factor: float = attr.field(default=1)
-    value: Union[float, int, str, None] = None
     # func: Union[
     #     None, Callable[[Tuple[int, ...]], str], Callable[[float], Any]
     # ] = attr.field(default=None)
     reg_value: Tuple[int, ...] = attr.field(init=False, factory=tuple)
+    on_change: Callable = attr.field(default=None)
```
Ivan-L, line 49: An `on_change` callable was introduced here, but change detection could be implemented in the add-on itself and the sunsynk library left clean.

hass-addon-sunsynk-dev/run.py
```diff
@@ -203,18 +239,36 @@ def setup_sensors() -> None:
...
-        STARTUP_SENSORS.extend(ALL_SENSORS[n] for n in startup_sens)
-    except KeyError as err:
-        _LOGGER.fatal("Invalid sensor as dependency - %s", err)
+    for sen_id, deps in sens_dependencies.items():
+        try:
+            sen = ALL_SENSORS.get(sen_id)
+        except KeyError as err:
+            _LOGGER.fatal("Invalid sensor as dependency - %s", err)
 
+        if sen_id not in sens and sen != RATED_POWER:  # Rated power does not change
```
Ivan-L, line 256: This will be true when the user does not add this sensor, which has dependants, to the add-on options.

hass-addon-sunsynk-dev/run.py
```diff
@@ -179,10 +202,23 @@ def startup() -> None:
...
     """Setup the sensors."""
     sens = {}
+    sens_dependencies: Dict[str, List[Sensor]] = defaultdict(list)
     startup_sens = {SERIAL.id, RATED_POWER.id}
+    filters: Dict[str, Filter] = {}
 
     msg: Dict[str, List[str]] = defaultdict(list)
 
+    def enqueue_hass_discovery_info_update(sen: Sensor, deps: List[Sensor]):
+        if not DEVICE:
```
Ivan-L, line 212: The `on_change` callables fire even on startup when sensor values are `None` and are read for the first time. This check prevents unnecessary queuing of discovery info updates.

hass-addon-sunsynk-dev/run.py
kellerza: :+1: maybe call it sens_dependants?

kellerza (commented):
some feedback, but maybe lets get the other part in first (or you are welcome to join the two if that makes rebasing easier)

hass-addon-sunsynk-dev/run.py
Ivan-L: 100%, I wanted to call it exactly that after my first commit but I had commented already and didn't want to push a new commit which would mess up the comments.

sunsynk/sensor.py
Ivan-L: Yes, that is exactly why I invoked the `on_change` callback (if set) when setting `value`. The `value` does get set when reading a new value (for example, me walking to the inverter and updating the value there), so that case is covered too.

I initially had the callback invoked inside the `update_value()` method, however that is overridden in subclasses and the callback ended up not being invoked. That is the exact reason why I invoke the callback in the setter of `value`.

Ivan-L:
Hi @kellerza I have rebased this PR on the latest changes. This is still in draft mode because there are two issues which have been brought to light which make me uncomfortable.

### Issue 1
The first is a pre-existing issue and is a minor one, but I thought I'd mention it nonetheless. In mqtt.py, there is an existing function which maps units of measurement to `device_class`.
```python
def hass_device_class(*, unit: str) -> str:
    """Get the HASS device_class from the unit."""
    return {
        "W": "power",
        "kW": "power",
        "kVA": "power",
        "V": "voltage",
        "kWh": "energy",
        "kVa": "energy",
        "A": "current",
        "°C": "temperature",
        "%": "battery",
    }.get(unit, "")
```
Any sensor with a `%` unit of measurement gets allocated the `battery` device class.
When opening the Home Assistant device that is created by the add-on, the HA UI picks the first sensor with a `battery` device class to use as the device SOC and if there are multiple such sensors, the first such sensor may be the wrong one. See here as an example:
image
This PR merely exposed this issue because it automatically adds the "Battery Low Capacity" sensor because the Prog1..6 Capacity writable sensors depend on it.

My suggestion is to only use a `device_class` of `battery` for the actual SOC of the battery, not for the configuration-type sensors, whether they are writable or not. One thought I had was to remove the `device_class` field on the `Entity` class in mqtt.py and rather add `device_class` on a new `SensorEntity` subclass of that. Then `RWSensor` would inherit from `Entity` and not have a `device_class` and the config-type battery capacity sensors would either be an instance of `RWSensor` (which would mean they are RW, but not editable because they are not quite ready to be editable yet) or an instance of `NumberEntity` if they are ready to be edited. For what it's worth, a base `Entity` class in this repo without a `device_class` field would align with the current HA [MqttEntity](https://github.com/home-assistant/core/blob/2497ff5a39292b7167733206fe2e1e63a6a3b86e/homeassistant/components/mqtt/mixins.py#L1038-L1066) implementation and a `SensorEntity` subclass which introduces `device_class` would align with the HA [MqttSensor](https://github.com/home-assistant/core/blob/2497ff5a39292b7167733206fe2e1e63a6a3b86e/homeassistant/components/mqtt/sensor.py#L363-L366) implementation. The `RWSensor` could have some default icons based on unit of measurement if we want to keep the battery (and other) icons that we got for free from HA based on `device_class`.

What are your thoughts around this? Do you have any other ideas on how to solve this? It's probably not the most important issue to tackle since it's purely a UI issue in Home Assistant, but I thought I'd mention it since this PR makes the issue more apparent. This can also be tackled as a separate PR.


### Issue 2
The second issue is introduced by this PR - if the user adds certain sensors (like the Prog1..6 Capacity ones) in OPTS _and_ those sensors have dependencies (which they do, on "Battery Low Capacity"), then "Battery Low Capacity" is automatically added to the list of sensors to poll. This was the only reliable way to ensure that this sensor is monitored for changes so that its dependents can be updated when there is a change. This is because of the current implementation of [publish_sensors](https://github.com/kellerza/sunsynk/blob/24697f84abdc135ac1fe7623d847fdd3ae6ee4ef/hass-addon-sunsynk-dev/run.py#L35-L52) in run.py where a `Filter` is required and the `force` parameter plays a role. If we can achieve the monitoring of a sensor for changes that was not explicitly added in OPTS in a different way, then that would negate the need to automatically add it in the proposed way in the current PR:
```python
if sen_id not in sens and sen != RATED_POWER:  # Rated power does not change
  fstr = suggested_filter(sen)
  msg[f"*{fstr}"].append(name)  # type: ignore
  filt = getfilter(fstr, sensor=sen)
  SENSORS.append(filt)
  sens[sen.id] = filt
  _LOGGER.info("Added sensor %s as other sensors depend on it", sen_id)
```
Any thoughts around this? If we try to monitor a sensor which has not explicitly added to OPTS and we do _not_ want to add it to the list of sensors added by the user, then `publish_sensors()` may need to be reworked somewhat and I may need your help/advice.

kellerza:
Hi @Ivan-L !

**On Issue 1** 

Leave device_class where it is, but remove setting a default in the base Entity class [here](https://github.com/kellerza/sunsynk/blob/main/hass-addon-sunsynk-dev/mqtt.py#L71-L73)
We can set defaults on the array of entities, before sensors are sent to `publish_discovery_info`. You can only set defaults on "" (make sure its not falsey, but ""), since you used None elsewhere...

Hope this makes sense... But at least for the battery target % you can set some other device class (or maybe None)

**On Issue 2** 

Maybe Entity needs a property like  `hidden`, in which case it will not `publish_discovery_info`.
You can still publish a hidden sensor's value to MQTT, so its easier to debug with mqtt explorer, but HASS wont pick it up since there is no discovery info for it.

Its faily easy to implement, simply add a filter [here](https://github.com/kellerza/sunsynk/blob/main/hass-addon-sunsynk-dev/mqtt.py#L231)

```python
entities = [e for e in entities if not e.hidden]
```

If hidden is false by default it should be fairly easy to set it true once you add it as a dependent

Ivan-L:
Hi @kellerza thanks for the tips!

I have implemented both suggestions, here are some extra comments.

### Issue 1
I removed the setting of the default device class as suggested. This in turn allowed me to remove a previous workaround for `NumberEntity` since its `device_class` field had different options from what the sensor entity has. I also took this opportunity to add some icons to the entities that get created from RWSensor since they do not use `device_class` for their appearance and the results look pleasing.

image

I added the unit to icon mapping to mqtt.py, but perhaps both the unit to device class and the unit to icon mappings belong in run.py - your call.

### Issue 2
Thanks for the suggestion. I implemented the filtering but one stack frame before the location that you suggested since we actually do not need to create entities out of hidden sensors. The value changes still get published via MQTT but there is no discovery info. I tested this and it works well - I walked to the inverter and changed a hidden sensor's value and the sensor's dependants in HA were updated with new min values.

I am marking this PR as ready for review.

hass-addon-sunsynk-dev/run.py
```diff
@@ -54,35 +64,85 @@ async def publish_sensors(sensors: List[Filter], *, force: bool = False) -> None
...
 async def hass_discover_sensors(serial: str, rated_power: float) -> None:
     """Discover all sensors."""
-    ents: List[Entity] = []
     dev = Device(
         identifiers=[OPT.sunsynk_id],
         name=f"Sunsynk Inverter {serial}",
         model=f"{int(rated_power/1000)}kW Inverter {serial}",
         manufacturer="Sunsynk",
     )
+    # pylint: disable=import-outside-toplevel
```
kellerza, line 73: do you need this?

hass-addon-sunsynk-dev/run.py
```diff
@@ -54,35 +64,85 @@ async def publish_sensors(sensors: List[Filter], *, force: bool = False) -> None
...
 
+        if isinstance(sensor, RWSensor):
+            ent.update(
+                {
+                    "entity_category": "config",
+                    "icon": hass_default_rw_icon(unit=sensor.unit),
+                }
+            )
+        else:
+            ent.update({"device_class": hass_device_class(unit=sensor.unit)})
```
kellerza, line 144: Rather assign directly to the dictionary ent when its just 1/2 properties

ent["device_class"] = ...

hass-addon-sunsynk-dev/run.py
```diff
@@ -107,6 +167,7 @@ def _handler(value):
             continue
 
         if isinstance(sensor, TimeRWSensor):
+            ent.update({"icon": "mdi:clock"})
```
kellerza, line 170: Here you can simply include icon in the parameters to SelectEntity a couple lines down...

hass-addon-sunsynk-dev/mqtt.py
```diff
@@ -63,14 +63,12 @@ class Entity:
...
+    icon: str = attr.field(default="")
 
     _path = ""
 
     def __attrs_post_init__(self) -> None:
         """Init the class."""
-        if not self.device_class:
-            # Try device_class from unit_of_measurement
-            self.device_class = hass_device_class(unit=self.unit_of_measurement)
         if not self.state_class and self.device_class == "energy":
```
kellerza, line 72: since we do not set the device_class, we might now miss total_increasing...

```suggestion
        if not self.state_class and self.unit_of_measurement == "kWh":
```

kellerza (commented):
Thanks @Ivan-L  this is looking very good!

hass-addon-sunsynk-dev/run.py
Ivan-L: Nope, can't remember where I copied and pasted it from. It's gone.

hass-addon-sunsynk-dev/run.py
Ivan-L: Unfortunately not because `icon` is already set on `ent` in the `RWSensor` branch and I got an error stating that I am supplying `icon` twice - once with the `ent` dictionary and once as an explicit field when I tried.

hass-addon-sunsynk-dev/mqtt.py
Ivan-L: I thought about this but we are still setting `device_class` for the regular (non-RW) sensors, so by the time `__attrs_post_init__` runs, there would be a `device_class` set?

Ivan-L: I double checked, the `device_class` and `state_class` fields still get set for the non-RW sensors:
image

kellerza: Thx for checking!

kellerza:
Thanks @Ivan-L these R/W sensors is a great addition!! 🐬🎉

PR #57, opened 2022-08-28
More writable sensors

### LINKED ISSUE #37: Enhancement: Writable sensors

Ivan-L:
Resolves #37 

This is the final PR dealing with writable sensors in #37.

The salient changes here are:

  - Introduced new `Battery Shutdown voltage`, `Battery Restart voltage` and `Battery Low Voltage` sensors
  - Introduced new `Float voltage` sensor upon which the `Prog1-6 Voltage` sensors depend for their `max` value
  - Made `Prog1-6 Charge` and `Prog1-6 Voltage` sensors writable
    image
  - Made `Equalization voltage` and `Absorption voltage` writable
  - Made `Battery Shutdown Capacity`, `Battery Restart Capacity` and `Battery Low Capacity` writable
  - Corrected `Prog1-6 Voltage` factor to be 0.01 (this was previously 0.1 which was incorrect)

If the `definitions.py` file can be formatted better, please do let me know.

My one wish from the Home Assistant [MQTT Number integration](https://www.home-assistant.io/integrations/number.mqtt/) is for it to support the `mode` configuration variable of [Number Entity](https://developers.home-assistant.io/docs/core/entity/number/) so that the voltage entities can be set to `box`. Currently the unchangeable `auto` value causes a `slider` to be used which is not wrong, but when dealing with voltage values I think the more correct control to use is `box`. I will  file a request for the MQTT Number integration to gain a `mode` configuration variable.

image

### DIFF
--- a/sunsynk/definitions.py
+++ b/sunsynk/definitions.py
@@ -6,7 +6,6 @@ from sunsynk.sensor import (
     InverterStateSensor,
     MathSensor,
     NumberRWSensor,
-    RWSensor,
     SDStatusSensor,
     SelectRWSensor,
     Sensor,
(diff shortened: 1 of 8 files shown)


sunsynk/definitions.py
```diff
@@ -146,18 +145,67 @@
...
+
+BATTERY_SHUTDOWN_CAPACITY = NumberRWSensor(217, "Battery Shutdown Capacity", "%")
+BATTERY_RESTART_CAPACITY = NumberRWSensor(218, "Battery Restart Capacity", "%")
+BATTERY_LOW_CAPACITY = NumberRWSensor(
+    219,
+    "Battery Low Capacity",
+    "%",
+    min=BATTERY_SHUTDOWN_CAPACITY,
+    max=BATTERY_RESTART_CAPACITY,
+)
```
Ivan-L, line 177: These sensors are currently called `Battery Shutdown Capacity`, `Battery Restart Capacity` and `Battery Low Capacity`. Similarly there are new `Battery Shutdown voltage`, `Battery Restart voltage` and `Battery Low voltage` sensors. I am wondering whether the `Battery` prefix should be removed since the Equalization, Absorption and Float voltage sensors do not have this prefix and perhaps there should be consistency. It would also read better here if the prefixes were gone, but would it be obvious that these sensors are battery-related?
image

sunsynk/definitions.py
```diff
@@ -146,18 +145,67 @@
...
     Sensor(611, "Bat1 Cycle"),
 )
-BATTERY_SHUTDOWN_CAPACITY = RWSensor(217, "Battery Shutdown Capacity", "%")
-BATTERY_RESTART_CAPACITY = RWSensor(218, "Battery Restart Capacity", "%")
-BATTERY_LOW_CAPACITY = RWSensor(219, "Battery Low Capacity", "%")
-_SENSORS += (BATTERY_SHUTDOWN_CAPACITY, BATTERY_RESTART_CAPACITY, BATTERY_LOW_CAPACITY)
+
+# Absolute min and max voltage based on Deye inverter
+MIN_VOLTAGE = 41
+MAX_VOLTAGE = 60
```
Ivan-L, line 157: These were obtained by checking out how far the values for the  Equalization, Absorption and Float voltage sensors can go on my Deye inverter.

sunsynk/definitions.py
```diff
@@ -179,6 +227,12 @@
 PROG5_TIME.max = PROG6_TIME
 PROG6_TIME.max = PROG1_TIME
 
+PROG_CHARGE_OPTIONS = {
+    4: "No Grid or Gen",
+    5: "Allow Grid",
+    6: "Allow Gen",
+    7: "Allow Grid & Gen",
+}
```
Ivan-L, line 235: The Sunsynk and Deye inverter UI have these as two checkboxes, but they operate on one register. I could try to have two switches operate on one register (Prog N Grid Charge, Prog N Gen Charge) but I haven't seen an example on how two different sensors can operate on the same register on the codebase.

kellerza: There are no other sensors that only use certain bits of a register today. Let's leave it like this for now

sunsynk/definitions.py
kellerza: I would rather add "Battery" to absorption and float than remove it from the rest.

kellerza (commented):
Lets update Battery absorption & battery float and then it should be good to merge!

sunsynk/definitions.py
Ivan-L: Done. I also added deprecated entries for `Equalization voltage` and `Absorption voltage` since they were present prior to the beta.

jabezz:
Hi guys

The work done here is absolutely amazing. Thanks!

I am interested in looking into adding the Work Mode related settings:

image

Where would I start looking at the register numbers and values?
I've come across the RS485 MODBUS communication protocol https://www.scss.tcd.ie/Brian.Coghlan/Elios4you/RS485_MODBUS-Hybrid-BACoghlan-201811228-1854.pdf

But the register values used in definitions.py are different.

Ivan-L:
Hi @jabezz it should be quite easy to add the new writable sensors, though there might be a few things to discuss prior to implementing, such as whether switching the above sensors on/off should cause other sensors to become enabled/disabled. Would you mind creating a new issue for the above sensors (Solar Export and Use Timer) so we can have the discussion there?

PR #58, opened 2022-09-03
Allow NumberEntity mode to be customized

Ivan-L:
As part of #37 the `NumberRWSensor` was introduced which mapped to the Home Assistant [MQTT Number Integration](https://www.home-assistant.io/integrations/number.mqtt/) entity. At the time, one could not set the `mode` of this entity so this resulted in the `mode` being `auto` for all `NumberRWSensor`s and nearly all of them got displayed as sliders. As of Home Assistant 2022.9.0, `mode` [can be customised](https://github.com/home-assistant/core/issues/77455) and this PR allows for the `mode` to be customised for `NumberRWSensor`s via an option in the add-on.

Before:
image

After (if setting the `mode` to `box`):
image
image

Note: If users do not have HA 2022.9.0, the `mode` field is ignored, therefore this is not a breaking change.

### DIFF
--- a/hass-addon-sunsynk-dev/CHANGELOG.md
+++ b/hass-addon-sunsynk-dev/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## **2022.09.16-0.2.5** - 2022-09-16
+
+- Sunsynk Dev Add-On
+
+  - Allow for customizing the `mode` of NumberEntity sensors allowing for forcing either `box` or `slider` mode instead of the default `auto` - requires Home Assistant 2022.9.0 for the effect to be seen - [#58](https://github.com/kellerza/sunsynk/pull/58)
+
 ## **2022.09.15-0.2.5** - 2022-09-15
 
 - Sunsynk Dev Add-On
--- a/hass-addon-sunsynk-dev/DOCS.md
+++ b/hass-addon-sunsynk-dev/DOCS.md
@@ -88,6 +88,10 @@
 
   A list of sensors to poll. You can use any sensor defined in the sunsynk Python library - [here](https://github.com/kellerza/sunsynk/blob/main/sunsynk/definitions.py).
 
+- `NUMBER_ENTITY_MODE`
+
+  When adding read/write sensors which present as number entities in Home Assistant, the default display mode is `auto`. This setting controls how the number entity should be displayed in the UI. Can be set to `box` or `slider` to force a display mode.
+
 - `MQTT_*`
 
   You will need a working MQTT sevrer since all values will be sent via MQTT.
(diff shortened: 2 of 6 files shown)


Ivan-L:
Hi @kellerza I am putting this PR back into draft mode. This morning I tried to decrease prog 2 capacity using box mode on my phone and struggled to do so because deleting characters forces the minimum value and adding characters forces the maximum. It's a quirk of the number input. I am starting to think that slider is good for mobile but box is good for desktop. Or perhaps using box mode for just voltage and amp units, not sure.

Your thoughts?

N1c084:
Hi 
Slider are ok on phone but not precise . Sometime the value appear with 2 decimal example 89.75% when auto update do the job
image
image

Maniac0Maniac:
Hi @Ivan-L

Been watching your progress for a while. Great job! It's turning into an awesome HA add-on. I vote Sliders for what it's worth - they look awesome!

Ivan-L:
Hi @N1c084 and @Maniac0Maniac 

Thank you for the feedback.

Regarding the imprecise sensor values (or values with many decimals) I believe this is because of the default filters that get applied to the sensors. We actually shouldn't use any filters for the RWSensors - I'll create a separate issue for this as it's not related to the slider - I get the same behaviour with the box inputs too.

After running with box mode for a few weeks, I must say that I prefer the sliders too because editing a value using box mode when the minimum and maximum is set is cumbersome. For example, say I want to modify `Prog 2 Capacity` and the current value is `70`% with the minimum being `30`% and maximum being `100`%. Let's say I want to change the value to `60`%. On my phone, if I backspace on the `70` deleting the `0` I am left with `7`, but `7` is out of range, so the value is coerced to become `30`%. If I delete the whole thing I am also left with `30`%. If I delete just the `3` I am left with `30`%. The solution is to select the first character and replace it with a `6` for it to become `60`%, but that is cumbersome.

So whilst I was the one advocating for `box` mode for all of the sensors and the Home Assistant MQTT integration maintainers kindly implemented the ability to customise the mode for the MQTT Number integration, I think leaving it as `auto` which results in most RW sensors using a slider, is actually fine.

What I can do is add a configuration option to the add-on in this PR so that people can experiment with it and have the choice of switching to `box` if desired, but I think the default of `auto` is fine for now.

Ivan-L:
@N1c084 please check whether there are still imprecise/average values when changing sensor values by installing the latest Dev version of the add-on. Another PR should have fixed this.

@kellerza I have updated this PR to, rather than force `box` mode for all `NumberRWSensor`s, keep the default of `auto` and allow for changing the `mode` through an option on the add-on. Would you be able to check and see if you're happy with the PR?

N1c084:
Hi 
image
Strange color in alarm

image

Error for prog2 not working now 
image

kellerza (approved):

kellerza:
🐬 Looks good, thanks @Ivan-L !

PR #62, opened 2022-09-15
Use round_robin filter for RWSensors

### LINKED ISSUE #61: Suggested filter for RWSensors should be round_robin
### Issue related to

Sunsynk Home Assistant Dev Add-On

### Describe the issue/bug

Currently if the user adds a sensor and does not explicitly choose a filter, the same set of filters get applied to read-only and writable sensors through the `suggested_filter` function. For example, for RW voltage sensors, an `avg` filter is used, which results in odd behaviour when the value is changed either on the inverter side or in the add-on itself in that when there is a change, the sensor value will gradually increase (with several decimal places) towards the new value but these intermediate average values are undesirable.

**Expected behavior**
It is proposed that the suggested filter for all RW sensors is `now` or `last`. I cannot think of a use case for a RW sensor where the filter should be different. Would there be any undesirable consequences of using the `now` or `last` filters for all RW sensors?

If there is agreement on the above, I will put up a PR.

Current implementation:
```python
def suggested_filter(sensor: Sensor) -> str:
    """Default filters."""
    if sensor.id.startswith("prog"):
        return "round_robin"
    f_id = {
        "battery_soc": "last",
        "fault": "round_robin",
        "grid_connected_status": "last",
        "overall_state": "step",
        "priority_mode": "round_robin",
        "sd_status": "step",
        "serial": "round_robin",
    }
    assert all(s in ALL_SENSORS for s in f_id)

    f_unit = {
        AMPS: "step",
        VOLT: "avg",
        WATT: "step",
        KWH: "last",
        CELSIUS: "avg",
    }
    res = f_id.get(sensor.id) or f_unit.get(sensor.unit) or "step"
    _LOGGER.debug("%s unit:%s, id:%s", res, sensor.unit, sensor.id)
    return res
```

Proposed change:
```python
def suggested_filter(sensor: Sensor) -> str:
    """Default filters."""
    if isinstance(sensor, RWSensor):
        return "now" # or "last"
    f_id = {
        "battery_soc": "last",
        "fault": "round_robin",
        "grid_connected_status": "last",
        "overall_state": "step",
        "sd_status": "step",
        "serial": "round_robin",
    }
    assert all(s in ALL_SENSORS for s in f_id)

    f_unit = {
        AMPS: "step",
        VOLT: "avg",
        WATT: "step",
        KWH: "last",
        CELSIUS: "avg",
    }
    res = f_id.get(sensor.id) or f_unit.get(sensor.unit) or "step"
    _LOGGER.debug("%s unit:%s, id:%s", res, sensor.unit, sensor.id)
    return res
```

Ivan-L:
Fixes #61

### DIFF
--- a/hass-addon-sunsynk-dev/CHANGELOG.md
+++ b/hass-addon-sunsynk-dev/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## **2022.09.16-0.2.5** - 2022-09-16
+
+- Sunsynk Dev Add-On
+
+  - Suggested filter for all RWSensors is now round_robin [#61](https://github.com/kellerza/sunsynk/issues/61)
+  - Filters now also get updated when force publishing sensor values
+
 ## **2022.08.30-0.2.5** - 2022-08-30
 
 - Python sunsynk module 0.2.5:
--- a/hass-addon-sunsynk-dev/config.yaml
+++ b/hass-addon-sunsynk-dev/config.yaml
@@ -1,6 +1,6 @@
 ---
 name: Sunsynk Inverter Add-on (dev)
-version: 2022.08.30-0.2.5
+version: 2022.09.16-0.2.5
 slug: hass-addon-sunsynk-dev
 description: Add-on for the Sunsynk Inverter
 startup: services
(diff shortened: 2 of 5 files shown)


kellerza (approved):
🚀  thanks @Ivan-L !

PR #80, opened 2023-01-10
Fix prog charge options

jacekowski:
Currently used values for PROG_CHARGE_OPTIONS set bit 2 high which does not seem to be correct (sunsynk app only uses bits 0 (grid charge) and 1 (gen charge))


kellerza:
Thanks @jacekowski - I started with #81 which should address your issue about bit 2.

Other inverters, including my Sunsynk, use bit 2 as well.

Issue #63, opened 2022-09-25
Compatibility with Sunsynk 12kw 3-phase inverter?


lubster85:
Hi there

I recently had a Sunsynk 12kw 3-phase inverter installed, but I'm a little unsure if this integration works with this model.
I purchased a USB-RS485 cable from Solar-Assistant (https://solar-assistant.io/shop/products/sunsynk_rs485) and tried starting up the integration.

I've tried a few ports on the inverter with no success yet, but I was hoping to sense-check myself here as I don't recall the 3-phase inverter being included in the list of tested inverters.

![IMG_1743](https://user-images.githubusercontent.com/47302755/192143754-d2a43a76-ff7b-4a7f-801f-619274329656.jpg)

My understanding is that the "Meter-485" port could be used for monitoring and also the "BMS" port.

Referencing Solar Assistant, they claim there was a firmware change recently.
> The inverter manufacturer released a firmware update which changed the port to read the inverter via RS485. An RJ45 splitter can be used to allow SolarAssistant to read the inverter while the inverter is also reading a battery via CAN.

Is there perhaps something else I may be missing or should try? 
If not, I'll try find a splitter and confirm if that makes a difference.

```
[s6-init] making user provided files available at /var/run/s6/etc...exited 0.
[s6-init] ensuring user provided files have correct perms...exited 0.
[fix-attrs.d] applying ownership & permissions fixes...
[fix-attrs.d] done.
[cont-init.d] executing container initialization scripts...
[cont-init.d] done.
[services.d] starting services
[services.d] done.
2022-09-24 08:45:31,191 INFO    Loading HASS OS configuration
2022-09-24 08:45:31,264 INFO    Filter *last used for battery_soc, total_battery_charge, total_battery_discharge, total_grid_export, total_grid_import, total_pv_power
2022-09-24 08:45:31,273 INFO    Protocol made connection.
2022-09-24 08:45:31,274 INFO    Connected to /dev/ttyUSB0
2022-09-24 08:45:41,286 ERROR   Timeout reading: 
2022-09-24 08:45:41,287 INFO    ############################################################
2022-09-24 08:45:41,287 INFO    No response on the Modbus interface, try checking the wiring to the Inverter, the USB-to-RS485 converter, etc
2022-09-24 08:45:41,287 INFO    ############################################################
2022-09-24 08:45:41,288 CRITICAL This Add-On will terminate in 30 seconds, use the Supervisor Watchdog to restart automatically.
[...]

Issue #71, opened 2022-11-19
Any idea what the unlisted inverter faults mean


cods4:
My inverters are always showing some sort of fault on the fault sensor, but none of them are listed on the manual or anywhere else I can find.

Surely they have some meaning, does anyone know what they are?
The ones I have are F65 through to F73

louispires:
Also wondered the same thing.

Mine always shows these 3: 
![image](https://user-images.githubusercontent.com/8841644/204790411-ff9e3214-91ca-4c13-be00-5a77b37b12f0.png)

I have in the past seen an error on my Inverter, BMS Failure but the HA integration never changed...

kellerza:
No idea to be honest, I really don't use this sensor, my sensor shows: F66, F67, F68, F70, F71. 
Everything seems ok on the inverter though.

I've never tried to test the [FaultSensor](https://github.com/kellerza/sunsynk/blob/main/sunsynk/sensor.py#L386) so no idea if it even works, it's simply my interpretation of the Sunsynk Modbus document. Open for PRs/suggestions

kellerza:
It seems this was solved by removing register 107 from the sensor.

Changes might only be in the multi version

kellerza:
Please check and let me know if still an issue?

cods4:
Yeh I noticed they were gone, which I guess solves it.

Thanks Johann

PR #93, opened 2023-01-31
Initial version of a definitions file for the 3phase inverters

kababoom:
Not remembering last time I created a pull request so hope all went well..

As discussed in issue #63 here's a separate file with registeres for the 3 phase inverters. There are sensors removed and sensors added like the different phase values (credits to @archi for these). 

The program 'mode options' I'm not sure of, might be there's a difference between inverters in this menu. 
I can set the time, charge options, power and voltage/soc. Enable/disable global time-of use and set the battery max charge/discharge amps. Which was my goal. There are a couple of optional sensors, they work and someone might be interested in these. 

The 3 phase inverters don't have a SD Card but if the sensor is removed it crashes because of filter line 201 so I left it.
(btw, you should remove register 107 for fault sensor, it's 'corrected Ah' not a fault register)

The way I see it there are two ways of implementing, simple drop-down list selecting single-phase / 3-phase definitions file. Or automatically by reading register 0. This will give 3 for a hybrid inverter and 5 for a 3 phase hybrid inverter. Based on that value the 'correct' definitions file can be loaded. 

I will be able to add the drop-down option, it will work but I'm not a python programmer so probably will not be the correct way.

I'm using a Waveshare rs485-eth converter, and it's working great so far. It's cheap (around 25eur inc shipping) https://www.waveshare.com/product/rs232-to-eth-b.htm If you want I can take a screenshot of the settings page to add to the readme.

If there's anything wrong in the file, am forgetting or any other comments please let me know.

Thanks for creating this great plugin...


kellerza:
Thanks for the PR @kababoom !

I reverted `definitions.py` and will accept `definitions3ph.py` and the switch sensor

Register 107 might solve #71

bruceborrett:
Can anyone confirm how to use this definitions file? Do I have to create a local copy of the addon and rename definitions3ph.py to definitions.py?

bruceborrett:
So I cloned the repo and renamed the file as per my previous comment, but with debug turned on I can see its still requesting the same registers from the original definitions file :confused:

PR #101, opened 2023-02-23
Change grid charge current to be writeable

jacekowski:


kellerza:
Thanks!

This will only take effect once the library has been updated

PR #113, opened 2023-03-21
Update deployment-options.md

stevelup:
I would suggest changing the wiring colours to T568B as they are the colours almost universally used today, and any pre-made patch lead will be in those colours.

### DIFF
--- a/www/docs/guide/deployment-options.md
+++ b/www/docs/guide/deployment-options.md
@@ -27,13 +27,13 @@ If the Inverter is close to your server/SBC running Home Assistant, you can use 
 ## RS485 wiring
 RS485 requires a twisted pair, this works with CAT5 network cable and RJ-45 connectors.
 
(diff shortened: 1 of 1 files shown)


kellerza:
There are already some installs with 568A, and a photo if I’m not mistaken. So ideally I don’t want to remove it, but rather keep it for reference

What do you think about adding 568B to the table?

stevelup:
Good idea - have both tables, then in the text you could say that a normal off the shelf ethernet cable is usually T568B. 

I've helped three other people set up your excellent integration, and all three of them messed up the same thing (used the T568A colours) - I accept three people is not exactly proof of a pattern, but it is a 100% hit in my experience!

I suspect most people will chop the end off a ready made patch lead when doing this, so they're more likely to be B than A.

kellerza:
Feeding an RJ45 connector through sprague, glands or conduit is no fun, so I only use wire and crimp my own 😉 

Also feeding 10+m through the inverter is not feasible when trying to keep the one end. But everyone might not have a crimping tool, so good to know there are 2 standards.

kellerza:
Closing in favor of #117

PR #115, opened 2023-03-21
Create Example of Turbo Energy system work mode control panel

cc120689:


kellerza:
Can you please update this PR to commit to the `lovelace-settings.md` file, you can add it to the end.

See the comment in #116 for more details

PR #117, opened 2023-03-22
Guide: add wiring pinouts for T568B

bmerry:
All my Ethernet cables seem to use T568B rather than T568A, so I added the pinout to help people who might otherwise just assume they have T568A without checking.


kellerza:
Thanks!

PR #118, opened 2023-03-23
Add both T568A and T568B to the RS485 wiring.

stevelup:
As per the earlier discussion.


stevelup:
Oops, sorry about that...

PR #141, opened 2023-05-23
Sunsynk Ecco combined port and USR-W610

Fr3d:
Sunsynk's new "Ecco" hybrid inverters have a combined RS485/CAN-BUS port instead of two distinct ports. This port is already used if you have a battery. Hopefully adding info about it to the documentation will save other people some time.

And I've been using/testing the noticeably cheaper USR-W610 RS485 WiFi gateway for weeks now and it's been flawless, so I think it can be added alongside the W630.


kellerza:
Thanks @Fr3d !

PR #144, opened 2023-06-02
SolarmanSunsynk Implementation

Xhanti:


kellerza:
```
pip install pre-commit
pre-commit install
```

will help with most of the lint issues

Xhanti:
Thanks for the tips @kellerza. 
I'll fix the formatting and other issues. 
Trying to test locally (ran out of time yesterday).

Xhanti:
@kellerza  got the addon onto my homeassitant instance but running into this issue, any ideas?

`[s6-init] making user provided files available at /var/run/s6/etc...exited 0.
[s6-init] ensuring user provided files have correct perms...exited 0.
[fix-attrs.d] applying ownership & permissions fixes...
[fix-attrs.d] done.
[cont-init.d] executing container initialization scripts...
[cont-init.d] done.
[services.d] starting services
[services.d] done.
Traceback (most recent call last):
  File "/usr/src/app/./run.py", line 10, in <module>
    from filter import RROBIN, Filter, getfilter
  File "/usr/src/app/filter.py", line 9, in <module>
    from sunsynk.definitions import AMPS, CELSIUS, KWH, SENSORS, VOLT, WATT, Sensor
ModuleNotFoundError: No module named 'sunsynk.definitions'
[cmd] ./run.py exited 1
[cont-finish.d] executing container finish scripts...
[cont-finish.d] done.
[s6-finish] waiting for services.
[s6-finish] sending all processes the TERM signal.`

kellerza:
If looks like you are not installing sunsynk in your Dockerfile

The `scripts\copy2local.cmd` scrip does 3 things:
- copy your version of the sunsynk library
- comments out pip install sunsynk
- uncomments a local install:
   
    ```Dockerfile
    # local test - install sunsynk
    RUN pip3 install -e sunsynk[pymodbus,umodbus]
    ```

Xhanti:
@kellerza 
 Thank you for the insights about copy2local, got everything working for testing. Its running now and I'm testing out the integration. 
 
 [s6-init] making user provided files available at /var/run/s6/etc...exited 0.
[s6-init] ensuring user provided files have correct perms...exited 0.
[fix-attrs.d] applying ownership & permissions fixes...
[fix-attrs.d] done.
[cont-init.d] executing container initialization scripts...
[cont-init.d] done.
[services.d] starting services
[services.d] done.
2023-06-05 21:09:30,263 INFO    options Loading HASS OS configuration
2023-06-05 21:09:30,280 INFO    Using Single phase sensor definitions.
2023-06-05 21:09:30,281 INFO    Connecting to tcp://00.00.00.00:0000
2023-06-05 21:09:30,288 INFO    Reading startup sensors rated_power, serial
2023-06-05 21:09:40,437 INFO    ############################################################
2023-06-05 21:09:40,437 INFO                Inverter serial number '0000000000000'
2023-06-05 21:09:40,437 INFO    ############################################################
2023-06-05 21:09:40,439 INFO    MQTT: Connecting to hass@core-mosquitto:1883
2023-06-05 21:09:40,452 INFO    MQTT: Connection successful
2023-06-05 21:09:50,861 ERROR   frame error reading registers starting at: 78
2023-06-05 21:09:56,110 ERROR   Caught error: V5 frame contains invalid sequence number
2023-06-05 21:09:56,714 ERROR   timeout reading register 78 (1)
2023-06-05 21:09:56,714 INFO    Retrying individual sensors: ['total_pv_energy', 'total_battery_charge', 'total_battery_discharge', 'total_grid_export', 'total_grid_import', 'battery_soc']
2023-06-05 21:10:11,664 ERROR   frame error reading registers starting at: 74
2023-06-05 21:10:16,904 ERROR   Caught error: V5 frame contains invalid sequence number
2023-06-05 21:10:17,547 ERROR   timeout reading register 74 (2)
2023-06-05 21:10:22,144 ERROR   frame error reading registers starting at: 81

Xhanti:
@kellerza so i've been playing with this for a bit and I think I'm quite a ways off from getting this to work. Implementing the base class and the changes in this PR can speak to a solarman dongle, but the dongle can't keep up with the number requests being made. In my local testing the dongle can handle one request every 10 seconds or so. Any faster and you start getting timeout errors and invalid responses. The home assistant addon right now assumes a physical connection and is very chatty over a physical link, it asks for registers much faster than once every 10 seconds. So to get this over the line I need to teach the addon to fetch as many registers as possible with as few calls as I can get away with. 

That sounds like alot of changes to how things work to get there, I'm trying to get to a minimal set of changes that can get me there, but its a work in progress. If you have any suggestions on how to go about this, that would be appreciated.

Failing that this project and the python library you have put together are fantastic, with everything I've learned in this PR I can put together a much less ambitious project that reads a few critical values I'm interested in and sends them to home assistant.

Let me know what you think :)

kellerza:
@Xhanti - I was considering updating the way the addon reads the registers, and even calculate filters.

If there was an option to read every 10 seconds (or maybe even more) would this work with your suggested driver?

Xhanti:
Hi @kellerza i have been playing with this since I last posted and things are a bit better I can see the light at the end of the tunnel. To answer your question, fetching all the registers every 10 seconds would definitely  get my driver over the line. I have a prototype that I can share with you that proves when fetching registers at that rate the solarman dongle can keep up with it and maybe do a bit better (query every 7-8 seconds).  


Having said that, I was able to get it reliably querying with things the way they are by changing this:
image
An allowed gap  of one effectively turned off grouping (made groups of one item) with the registers I wanted, this ended up making many little calls and overwhelming the dongle. I increased that to 10 which allowed the max_group_size to be 60 (the defalut). I got very good perf after that :). 

image

So now I get values and consistently, but there is still a bug. Instead of waiting 10 seconds between polls, it waits 10 min. 
My python is quite rusty so I'm trying to figure out what is causing the 10 min delay. Otherwise the driver looks like its in good shape.

Xhanti:
@kellerza this is all working and tested my side. with these changes one can reliably get  data out of a solarman dongle.
they are 2 things to take note of:

1. The more registers you ask for, the more likely you are to overwhelm the dongle. It can keep up with 10 registers easy. For my testing I asked for these sensors total_pv_energy,total_battery_charge,total_battery_discharge,total_grid_export,total_grid_import,battery_soc and essential_power  Batching helps mitigate this quite a bit by making fewer but bigger requests, the dongle seems to like that
2. When you query the solarman dingle as frequently as every second, you start getting responses that are meant to go to the official solarman cloud instance. The pysolarman library knows about this case and deals with it, so does the implementation I provided (see the retry mechanism). This means that you will see data in home assistant, but not the solarman app (it will say the dongle is online, but that inverter is offline. presumably because it gets no message from the inverter). Querying every 10 seconds allows the frames meant for official solarman cloud instance to get through and then home assistant and the solarman app can work side by side. 

Please take a look at the changes and see if you happy with them. Thanks
image

jmccrohan:
@Xhanti I just stumbled across this project and this PR in particular and it is very timely.. I have written a Modbus RTU over TCP to Solarman Proxy that I need to package up as a HA addon. I've just tidied it up and published it [here](https://github.com/jmccrohan/pysolarmanv5/blob/main/examples/solarman_proxy.py). This allows you to use the native Modbus integration within HA to communicate with the Addon, which will then encapsulate/decapsulate the requests to/from the Solarman logging stick. Similar idea to what you are trying to achieve here.

hass-addon-sunsynk-multi/config.yaml
```diff
@@ -44,11 +45,12 @@ options:
   DEBUG: 0
   DEBUG_DEVICE: "/dev/ttyAMA0"
 schema:
-  DRIVER: list(pymodbus|umodbus)
+  DRIVER: list(pymodbus|umodbus|solarman)
   INVERTERS:
     - SERIAL_NR: str
       HA_PREFIX: str
       MODBUS_ID: int(1,16)
+      DONGLE_SERIAL_NUMBER: str
```
kellerza, line 53: This is [optional](https://developers.home-assistant.io/docs/add-ons/configuration/#options--schema)

```suggestion
      DONGLE_SERIAL_NUMBER: str?
```

hass-addon-sunsynk-multi/options.py
```diff
@@ -20,13 +20,18 @@ class InverterOptions:
     modbus_id: int = 0
     ha_prefix: str = ""
     serial_nr: str = ""
+    dongle_serial_number: str = ""
 
     @classmethod
     def factory(cls, opt: dict) -> InverterOptions:
         """Create a class from the options."""
         modbus_id = int(opt.pop("MODBUS_ID"))
+        dongle_serial_number = opt.pop("DONGLE_SERIAL_NUMBER")
```
kellerza, line 29: Since it's optional, test if it exists, or wrap in an try/catch with Indexerror

hass-addon-sunsynk-multi/run.py
```diff
@@ -126,6 +133,7 @@ def setup_driver() -> None:
             "server_id": opt.modbus_id,
             "timeout": OPT.timeout,
             "read_sensors_batch_size": OPT.read_sensors_batch_size,
+            "dongle_serial_number": opt.dongle_serial_number,
```
kellerza, line 136: Should you not add this only if its the solarman driver?

or create a driver_specific dict (say on line 105) that you merge here? Then you can add solarmansunsynk speciifc options where you assign the factory var on line 121

```suggestion
            **driver_specific_options
```

sunsynk/sunsynk.py
```diff
@@ -24,6 +25,8 @@ class Sunsynk:
     timeout: int = attrs.field(default=10)
     read_sensors_batch_size: int = attrs.field(default=60)
     timeouts: int = 0
+    dongle_serial_number: int = attrs.field(default=0)
```
kellerza, line 28: dongle_serial_number should ideally move to `solarmansunsynk.py`, as it is specific to that class

sunsynk/sunsynk.py
```diff
@@ -24,6 +25,8 @@ class Sunsynk:
...
+    allow_gap: int = 1
```
kellerza, line 29: I like the addition of allow_gap, as having this at 1 is a bit counterintuitive to what we're trying to do with grouping them (sending less data). In your case you sending less data is important

kellerza (changes requested):
Thanks @Xhanti - this is looking good. Some small comments.

The main suggesiton is making the dongle serial number optional

kellerza:
> @Xhanti I just stumbled across this project and this PR in particular and it is very timely.. I have written a Modbus RTU over TCP to Solarman Proxy that I need to package up as a HA addon. I've just tidied it up and published it [here](https://github.com/jmccrohan/pysolarmanv5/blob/main/examples/solarman_proxy.py). This allows you to use the native Modbus integration within HA to communicate with the Addon, which will then encapsulate/decapsulate the requests to/from the Solarman logging stick. Similar idea to what you are trying to achieve here.

@jmccrohan interesting idea, so the sunsynk addon should be able to connect via this proxy?

In my view the native modbus integration in HA is a temporary solution. Once the particular modbus config is used by a critical number of users, it makes more sense to wrap it in a dedicated implementation/addon.

Having the register numbers in an addon is to abstract the inverter internals, else 10s/100s of users end up maintaining their own list of mappings. In this way everyone benefits from small changes/additions to sensors over time with simple updates. @Xhanti's idea of adding the pysolarman driver here works toward the same goal of making it really easy for people to get started.

hass-addon-sunsynk-multi/config.yaml
Xhanti: I've made this change

hass-addon-sunsynk-multi/options.py
Xhanti: I've made this change

hass-addon-sunsynk-multi/run.py
Xhanti: I've made this change

sunsynk/sunsynk.py
Xhanti: I've made this change

sunsynk/sunsynk.py
Xhanti: :)

Xhanti:
@jmccrohan I agree your knew project looks pretty interesting. I've found the pysolarman library very handy, but one of the goals I had here was to limit the amount of register interpretation I would have to do. The sensors abstraction (for example something like [this](https://github.com/kellerza/sunsynk/blob/f0bf315095f04d1f46827b92f5d0d85d045a585c/sunsynk/definitions.py#L89) is tricky to get right) in this libray was the main reason I put this PR together, I was already able to read the registers and put them somewhere with pysolarman :)

Xhanti:
@kellerza I've made all the changes I think I need to, anything outstanding here?

kellerza (approved):

kellerza:
> @kellerza I've made all the changes I think I need to, anything outstanding here?

Looking good, only some smaller pylint issues.

I’ll merge when I’m ready to release a new version. Since the addon gets built directly from github, importing an unknown class will cause an issue for new installs

Xhanti:
👍🏽

kellerza:
Thanks @Xhanti !

cods4:
Awesome work.

Just wondering, is this able to write settings to the inverter (or will the be possible in the future)? Or does the dongle not allow that?
I know Deye (and I assume Sunsynk) can allow ability to write settings through the dongle, but not sure if this requires credentials etc.

Xhanti:
It should allow writing as well

Xhanti:
Thanks @kellerza!

PR #197, opened 2023-10-18
Added battery max charge and discharge sensors

iBeech:
Allows end user to monitor if their charge / discharge rates are being reduced.

@kellerza I'm unsure how to test these changes.

### DIFF
--- a/src/sunsynk/definitions.py
+++ b/src/sunsynk/definitions.py
@@ -31,8 +31,8 @@ SENSORS += (
     Sensor(190, "Battery power", WATT, -1),
     Sensor(191, "Battery current", AMPS, -0.01),
     # Charge and Discharge limit vary based on temperature and SoC
-    Sensor(314, "Battery Charge Current Limit", AMPS, -0.01),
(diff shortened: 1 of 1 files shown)


kellerza:
The best way to test new sensor definitions is to use [custom sensors](https://kellerza.github.io/sunsynk/reference/mysensors)

iBeech:
OK, tested- for some reason, these particular sensors do not work. They just return 0A.

iBeech:
@kellerza resolved. I  was targeting the wrong sensors. These new sensors correctly retrieve the appropriate information. Also resolved some unit test issues

kellerza:
Thanks @iBeech !

ap20132022:
> OK, tested- for some reason, these particular sensors do not work. They just return 0A.

Hi, are you sure, that it works. On the manual, I found that:
210 Max A Charge R/W [0,185] 1A 0-185A
211 Max A discharge R/W [0,185] 1A 0-185A
I have a 12k sg04 (LV Inverter)

PR #233, opened 2024-02-07
update 3 phase hv definitions

kharnt0x:
Adding a bunch more registers including for battery 2 (this inverter has 2 separate battery ports/bms ports)

Also settings for enabled time of use by day

### DIFF
--- a/src/sunsynk/definitions3phhv.py
+++ b/src/sunsynk/definitions3phhv.py
@@ -10,7 +10,7 @@ from sunsynk.rwsensors import (
 )
 from sunsynk.sensors import (
     BinarySensor,
-    FaultSensor,
+    HVFaultSensor,
     InverterStateSensor,
     MathSensor,
     SDStatusSensor,
@@ -26,11 +26,16 @@ SENSORS = SensorDefinitions()
 # Battery
 ##########
 SENSORS += (
-    TempSensor(586, "Battery temperature", CELSIUS, 0.1),
-    Sensor(587, "Battery voltage", VOLT, 0.1),
-    Sensor(588, "Battery SOC", "%"),
-    Sensor(590, "Battery power", WATT, -10),
-    Sensor(591, "Battery current", AMPS, -0.01),
+    Sensor(586, "Battery 1 temperature", CELSIUS, 0.1),
+    Sensor(587, "Battery 1 voltage", VOLT, 0.1),
+    Sensor(588, "Battery 1 SOC", "%"),
+    Sensor(590, "Battery 1 power", WATT, -10),
+    Sensor(591, "Battery 1 current", AMPS, -0.01),
+    Sensor(589, "Battery 2 SOC", "%"), #correct
+    Sensor(593, "Battery 2 voltage", VOLT, 0.1), #correct
+    Sensor(594, "Battery 2 current", AMPS, -0.01), #correct
+    Sensor(595, "Battery 2 power", WATT, -10), #correct
+    Sensor(596, "Battery 2 temperature", CELSIUS, 0.1), #correct I think
 )
 
 #################
@@ -74,6 +79,9 @@ SENSORS += (
     Sensor(650, "Load L1 power", WATT, -1),
     Sensor(651, "Load L2 power", WATT, -1),
     Sensor(652, "Load L3 power", WATT, -1),
+    Sensor(644, "Load L1 voltage", VOLT, 0.1),
+    Sensor(645, "Load L2 voltage", VOLT, 0.1),
+    Sensor(646, "Load L3 voltage", VOLT, 0.1),
 )
 
 #
@@ -81,7 +89,7 @@ SENSORS += (
 # Solar Power
 ################
 SENSORS += (
-    MathSensor((672, 673), "PV power", WATT, factors=(10, 10)),  # pv1 power,pv2 power
+    MathSensor((672, 673, 674, 675), "PV power", WATT, factors=(10, 10, 10, 10)),  # pv1 power,pv2 power
 )
 ################
 # MPPT 1 Solar Power
@@ -161,7 +169,7 @@ SENSORS += (
     Sensor(
         0, "Device Type"
     ),  # {2: "Inverter", 3: "Hybrid Inverter", 4: "Micro Inverter", 5: "3 Phase Hybrid Inverter" }),
-    FaultSensor((555, 556, 557, 558), "Fault"),
+    HVFaultSensor((555, 556, 557, 558), "Fault"),
     InverterStateSensor(500, "Overall state"),
     SDStatusSensor(0, "SD Status", ""),  # type: ignore
     SerialSensor((3, 4, 5, 6, 7), "Serial"),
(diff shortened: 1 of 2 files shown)


src/sunsynk/definitions3phhv.py
```diff
@@ -266,7 +283,7 @@
 SENSORS += SelectRWSensor(
     142,
     "Load Limit",
-    options={0: "Allow Export", 1: "Essentials", 2: "Zero Export"},
+    options={0: "Allow Export", 1: "Zero Export To Non Essential", 2: "Zero Export To Grid CT"},
```
rixxxx, line 313: Isn't it "Zero Export To Essential"?

src/sunsynk/definitions3phhv.py
```diff
@@ -181,12 +189,14 @@
 SENSORS += (
     NumberRWSensor(128, "Grid Charge Battery current", AMPS, max=210),
     SwitchRWSensor(130, "Grid Charge enabled"),
+    NumberRWSensor(127, "Grid Charge Start Battery SOC", "%"),
+    NumberRWSensor(128, "Grid Charge Amps", AMPS),
```
rixxxx, line 218: 128 is defined twice

src/sunsynk/definitions3phhv.py
kharnt0x: from my understanding it's zero export to whatever is between the grid terminals of the inverter and the CT (which should be at the grid connection). Essential is the stuff on the load terminals

rixxxx: My 20kW inverter has "Zero Export To Load" and "Zero Export to CT" in the menu. Current "Zero Export" is "Zero Export to CT" in the device vocabulary. So the other "Zero Export to Load" is not "Zero Export To **Non** Essential" but "Zero Export To Essential" in my understanding. In my opinion manufacturer's options are in this case pretty clear.

kharnt0x: yeah we can keep the manufacturer terminology to not be confusing. There are so many ways to configure the inverter and load management that my assumption can be wrong

src/sunsynk/definitions3phhv.py
```diff
@@ -399,8 +426,140 @@
...
+#################
+# Advanced Inverter Software Configuration Settings
+#################
+SENSORS += (
+    ### no idea why there is a "no work" option, but it's in the spec and for me, beep is set to No Work when I disabled in on the inverter...
+    SelectRWSensor(228, "Time synchronization", options={0: "No work", 1: "No work", 2: "Disable", 3: "Enable"}, bitmask=0x03),
+    SelectRWSensor(228, "Beep", options={0: "No work", 4: "No work", 8: "Disable", 12: "Enable"}, bitmask=0x0C), # mine was set to 4 after I disabled it on the inverter
+    SelectRWSensor(228, "AM/PM", options={0: "No work", 16: "No work", 32: "Disable", 48: "Enable"}, bitmask=0x30), #I guess this is 24 hour or AM/PM time display
+    SelectRWSensor(228, "Auto dim", options={0: "No work", 64: "No work", 128: "Disable", 192: "Enable"}, bitmask=0xC0),
+    SelectRWSensor(228, "Allow Remote", options={0: "No work", 16384: "No work", 32768: "Disable", 49152: "Enable"}, bitmask=0xC000), # mine was set to 32768 from factory (a "no work" option)
```
rixxxx, line 544: I have to admit, I'm not a fan of having only a bitmask and some meaningless value. It works but its ugly. @kellerza what do you think, should we implement shifting or use it like done here?

kellerza: You can use `2<<13` or even hex `0x4000`

src/sunsynk/sensors.py
```diff
@@ -183,6 +183,23 @@ def reg_to_value(self, regs: RegType) -> ValType:
             val += chr(b16 & 0xFF)
         return val
 
+@attrs.define(slots=True, eq=False)
+class EnumSensor(Sensor):
+    """Sensor with a set of enum values. Like a read-only SelectRWSensor"""
```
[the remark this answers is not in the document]
kellerza: IF you run the lint tools offline (black/pydocstyle etc), they will pick up these types of issues

src/sunsynk/definitions3phhv.py
rixxxx: Sure, but they are still hard to compare with the modbus manual where eg register 228 has 2 bit wide options with the values 00, 01, 10, 11. Where 10 = disable and 11 = enable and other 2 unused. There are more registers with the same pattern. Anyway, I'm not against it.

kellerza: You can also combine bniiary + shift

```python
>>> 0b10<<13
16384
>>> 2<<13
16384
```

kellerza: if you use the same shift value for the mask and values it should be clearer

```python
>>> bin(0xc000)
'0b1100000000000000'
>>> bin(0b11<<14)
'0b1100000000000000'
```

rixxxx: So the question is, do you prefer to have 49152 or 0xC000 or 0x11<<14 in the code?

kellerza: `0x11<<14` is different than `0b11<<14`

but if you have `<<14` for mask+values in the same line it should be fairly clear

rixxxx: Yes, my mistake, I meant 0b11<<14. And having the same shift for mask and value makes it readable enough. I have a patch in the queue waiting for LV test which needs the same.

kharnt0x:
I also added state_class of measurement for measurements... This was because in my own usage things like PV_power were not showing up as statistics properly in HA

src/sunsynk/definitions3phhv.py
kharnt0x: agree it's nice to have it more readable, I'll make the change

kharnt0x: done

src/sunsynk/sensors.py
```diff
@@ -215,3 +232,55 @@ def reg_to_value(self, regs: RegType) -> ValType:
...
+
+    def reg_to_value(self, regs: RegType) -> ValType:
+        """Decode HV Inverter faults."""
+        faults = {
+            1: "DC Inversed Failure",
+            7: "DC START Failure",
+            13: "Working mode change",
+            15: "AC over current SW Failure",
+            10: "Aux Power Board Failure",
+            13: "Working mode changed",
```
kellerza, line 248:
```suggestion
```

src/sunsynk/sensors.py
```diff
@@ -215,3 +232,55 @@ def reg_to_value(self, regs: RegType) -> ValType:
...
+        err = []
+        off = 0
+        for b16 in regs:
+            for bit in range(16):
+                msk = 1 << bit
+                if msk & b16:
+                    msg = f"F{bit+off+1:02} " + faults.get(off + msk, "")
+                    err.append(msg.strip())
+            off += 16
+        return ", ".join(err)
```
kellerza, line 286:
```suggestion
        return ", ".join(err)

```

PR #236, opened 2024-02-12
Grid Peak Shaving for 3PH HV and LV inverters

rixxxx:
Its tested on HV SUN-20K-SG01HP3-EU-AM2.

### DIFF
--- a/src/sunsynk/definitions3ph.py
+++ b/src/sunsynk/definitions3ph.py
@@ -176,16 +176,8 @@ SENSORS += (
     NumberRWSensor(109, "Battery Max Discharge current", AMPS, max=210),
     NumberRWSensor(102, "Battery Capacity current", AMPS, max=2000),
     NumberRWSensor(191, "Grid Peak Shaving power", WATT, max=100000),
-    SelectRWSensor(
-        178,
-        "Grid Peak Shaving",
-        options={
-            0: "No work",
-            0b01 << 4: "No work",
-            0b10 << 4: "Disable",
-            0b11 << 4: "Enable",
-        },
-        bitmask=0b11 << 4,
+    SwitchRWSensor(
+        178, "Grid Peak Shaving", off=0b10 << 4, on=0b11 << 4, bitmask=0b11 << 4
     ),
 )
 
--- a/src/sunsynk/definitions3ph.py
+++ b/src/sunsynk/definitions3ph.py
@@ -175,6 +175,8 @@ SENSORS += (
     NumberRWSensor(108, "Battery Max Charge current", AMPS, max=210),
     NumberRWSensor(109, "Battery Max Discharge current", AMPS, max=210),
     NumberRWSensor(102, "Battery Capacity current", AMPS, max=2000),
+    NumberRWSensor(191, "Grid Peak Shaving power", WATT),
+    SelectRWSensor(178, "Grid Peak Shaving", options={0b10<<4: "Disable", 0b11<<4: "Enable"}, bitmask=0b11<<4),
 )
 
 # Additional optional sensors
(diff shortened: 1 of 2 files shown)


src/sunsynk/definitions3ph.py
```diff
@@ -175,6 +175,13 @@
     NumberRWSensor(108, "Battery Max Charge current", AMPS, max=210),
     NumberRWSensor(109, "Battery Max Discharge current", AMPS, max=210),
     NumberRWSensor(102, "Battery Capacity current", AMPS, max=2000),
+    NumberRWSensor(191, "Grid Peak Shaving power", WATT),
+    SelectRWSensor(
```
kellerza, line 179: If these are enable&disable only, can you please use a BinarySensor

rixxxx: By the documentation these are actually
 00: "No work"
 01: "No work"
 10: "Disable"
 11: "Enable"
like many bitfield registers in those inverters.

I can add those "No work" entries like it was done in #233 or make it BinarySensor. I initially had all 4 entries but removed "No work"s but forgot the change the type.

rixxxx: probably makes sense to keep them same

kellerza: This will *no work*. If you have two entries with the same name, how would you choose which register value to use if you select a name in the frontend? So you will always end up with one of them.

What is "No work"? Is it not just some bad translation of "Not working" / "Not used"?

Added a test case to have both on&off parameters to a Binary sensor, which is still my preference.

rixxxx: I now looked into the code and it seems BinarySensor is read-only. This is read-write register to control Grid Peak Shaving. One of the imlemented registers controls the peak shaving power and the other enables/disables the feature.

rixxxx: Current code has 4 options, original had only Enable and Disable
image

In the UI it looks nicer with only Enable and Disable present.

kellerza: Did you have a look at `SwitchRWSensor`? This has an on & off option.

I'll repeat the earlier "No work" comment. Having 2 options with the same text value makes no sense and you can never be sure which register value you selected.

rixxxx: Switched over to SwitchRWSensor

rixxxx: I'll remove my pull request, it seems it was already implemented as part of commited #233 

    SelectRWSensor(
        178,
        "Grid peak-shaving",
        options={0b10 << 4: "Disable", 0b11 << 4: "Enable"},
        bitmask=0b11 << 4,
    ),

rixxxx: I missed it too. It had multiple iterations and I lost track at one point ;) Works for me, however I started to like your suggestion of using SwitchRWSensor. So why not convert all those Enable/Disable selectors to switches?

kellerza: @kharnt0x - any objection to move register 178 to Switches instead of the dropdowns with enable/disable?

Impact would be that their name will change in HA to `switch.xxx` & any automations (if you already have any)

rixxxx: Speaking of impact. I just noticed my automations broke:) Part of #233 
 PROG_CHARGE_OPTIONS = {
     0: "Off",
-    1: "On",
+    1: "Grid Only",
+    2: "Gen Only",
+    3: "Grid and Gen",
 }

rixxxx: Making things as much as possible as switches makes lot of sense in the long run.

PR #250, opened 2024-02-26
Main

RhysMcW:
Prog{x} Time select list kellerza#246


src/ha_addon_sunsynk_multi/a_sensor.py
```diff
@@ -189,7 +189,7 @@ def on_change(val: float | int | str | bool) -> None:
             ent["icon"] = "mdi:clock"
             self.entity = SelectEntity(
                 **ent,
-                options=sensor.available_values(15, ist.get_state),
+                options=sensor.available_values(OPT.number_entity_mode, ist.get_state),
```
kellerza, line 192: Should this not be `OPT.prog_time_interval`?

kellerza (commented):
Thanks!

src/ha_addon_sunsynk_multi/a_sensor.py
RhysMcW: yep, you are correct

kellerza:
Thanks @RhysMcW !

I made the ocnfig key optional & shortened the list. Hopefully everyone can update without any missing config keys

Issue #235, opened 2024-02-12
sync 3PH LV and HV definitions


rixxxx:
### Describe the issue/bug

There has been recent development in the 3PH HV definition which apply to LV aswell. Both #218 and #233 add things which are only present in HV definition.

### Expected behavior

Definitions should differ only by multiplyers.
Does it make sense to try to merge LV and HV into one and have conditions for multiplyers?

Issue #201, opened 2023-10-31
Deye-12k-SG04LP3-EU 3ph inverter - error handling Low Capacity, use_timer and progX_charge


peter-mul:
### Your Inverter 
Deye-12k-SG04LP3-EU, Firmware 1140
This is a 3ph inverter

### Describe the issue/bug and what you expect
1. reading battery Low Capacity from the inverter is correct in homeassistant.  
Writing (using the slider) a value in HA sets the value to 0 in the inverter
The expectation is that wenn I select a value like i.e. 50 in HA this value is stored/written to the inverter.

2. Unchecked Use Timer checkbox in the inverter is not read correctly.
In the log we see: use_timer: Unknown register value 254
The value 254 is read, when the checkbox in the inverter ist unchecked.
So the expectation is that this value is recognized as off
currently the code shows: SwitchRWSensor(146, "Use Timer", on=255)
maybe we need something like SwitchRWSensor(146, "Use Timer", off=254, on=255)

3. The 3ph inverter has for the prog charge options two checkboxes Grid Charge and Gen Charge.
In the single phase inverter it is already coded to support the two options, so I expect it has also these options.
Currently the 3ph implementation only handles 0 and 1, one bit of the bitmap and for the second bit we get errors in the log
Log: 17:12:15 WARNING prog1_charge: Unknown register value 3
Log: 17:12:15 WARNING prog3_charge: Unknown register value 2
The expectation is that the second checkbox for Gen Charge is handled too, probably just like in the 1ph inverter.
PROG_CHARGE_OPTIONS = {
    0: "No Grid or Gen",
    1: "Allow Grid",
    2: "Allow Gen",
    3: "Allow Grid & Gen",
}

Thanks for the add-on!

kellerza:
On 2.

Try this definition with mysensors.py -- see the docs. 

```python
SwitchRWSensor(146, "Use Timer", bitmask=1)
```

On 1&3 you can experiment with the 1ph definitions on you 3ph definitions, just use the 1-ph definition in mysensors.py

If you do manage to fix these, please consider opening a PR to modify the 3ph definitions!

PR #253, opened 2024-02-29
Merge 3 phase LV&HV definitions

kellerza:
There seems to be a big overlap in the current 3PH sensor definitions.

This PR creates a common part, extended by LV&HV specific sensors. Although I'm starting to doubt if they should be split at all?

Any concerns? @rixxxx  @kharnt0x 

Solves #235 #201 (1&2)


kellerza:
Thx @rixxxx - I added a test to compare old vs new.

This test also shows sensors where the definitions changed slightly to align the two

<https://github.com/kellerza/sunsynk/pull/253/files#diff-ed03843640b2cef70980638e0efff27d7be035aad065b52aaa0c7bd70a8db0e8R18>

PR #258, opened 2024-03-11
Fix battery temperature sensor

rixxxx:
Use correct sensor type to get correct value. Requires offset like other temperature registers.


kellerza:
We should probably get rid of the old_lv/hv completely, as test will keep failing every time we make a change

rixxxx:
Probably. Throwing only one error at a time doesn't help either :)

PR #274, opened 2024-04-21
Merge few definitions from HV to common and from common to LV

rixxxx:
1. Move unnumbered battery sensors to LV definition, currently they get picked up by eg power_flow_card package and produce additonal sensors with wrong values.
2. Replace common Grid Connected with the one from HV definition. Sensor 194 is probably inherited from 1PH definition and doesn't work.
3. Deprecate LV Grid Conneced Status and move HV version of it to common.
4. Remove HV Rated power and ProgX power definitions. They are the same what is already in common.

Tested on HV. Partly tested on LV.


rixxxx:
Positive test results on 3PH LV too.

kellerza:
Thanks, can you remove the test_migration test cases?

It does not make sense to try keep these around&fix them

rixxxx:
Sure

PR #349, opened 2024-10-08
Add env vars support to options

maslyankov:
Added support of env vars for docker image

### DIFF
--- a/hass-addon-sunsynk-multi/README.md
+++ b/hass-addon-sunsynk-multi/README.md
@@ -10,3 +10,11 @@ See the docs for more info - [https://kellerza.github.io/sunsynk](https://keller
 Example of Home Assistant Energy Management, enabled by this addon.
 
 ![HASS Energy management](https://github.com/kellerza/sunsynk/raw/main/images/energy.png)
+
+## Usage of env vars
+
+You can also use env vars for config options. For them to work, you need to add one additional env var**: 
+
+S6_KEEP_ENV: 1
+
+** More info here: https://github.com/just-containers/s6-overlay#customizing-s6-overlay-behaviour
--- a/src/ha_addon_sunsynk_multi/options.py
+++ b/src/ha_addon_sunsynk_multi/options.py
@@ -15,24 +15,36 @@ from ha_addon_sunsynk_multi.timer_schedule import Schedule
 _LOGGER = logging.getLogger(__name__)
 
 
-def unmarshal(target: object, json: dict) -> object:
-    """Update options."""
-    if not isinstance(json, dict):
-        _LOGGER.error("invalid options %s", json)
+def unmarshal(target: object, json_data: dict) -> object:
+    """Update options from a JSON dictionary."""
+    if not isinstance(json_data, dict):
+        _LOGGER.error("Invalid options format: %s", json_data)
+        return target
+
     _lst = getattr(target, "_LISTS", {})
-    for key, val in json.items():
+
+    for key, val in json_data.items():
         key = fixkey(key)
-        if key in _lst:
+        
+        if key in _lst:  # Handle lists (e.g., inverters, schedules)
             newcls = _lst[key]
-            newv = [unmarshal(newcls(), item) for item in val]
-            setattr(target, key, newv)
+            if isinstance(val, list):
+                # Preserve existing values if already set (e.g., from environment)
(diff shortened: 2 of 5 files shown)


hass-addon-sunsynk-multi/README.md
```diff
@@ -10,3 +10,11 @@ See the docs for more info - [https://kellerza.github.io/sunsynk](https://keller
 Example of Home Assistant Energy Management, enabled by this addon.
 
 ![HASS Energy management](https://github.com/kellerza/sunsynk/raw/main/images/energy.png)
+
+## Usage of env vars
+
+You can also use env vars for config options. For them to work, you need to add one additional env var**: 
+
+S6_KEEP_ENV: 1
```
kellerza, line 18: Can you use env vars with addons?

Or would this be for the standalone deployments? might be best to document [here](https://kellerza.github.io/sunsynk/guide/standalone-deployment)

src/ha_addon_sunsynk_multi/options.py
```diff
@@ -52,24 +53,24 @@ class Options:
...
     _LISTS = {"inverters": InverterOptions, "schedules": Schedule}
 
-    mqtt_host: str = ""
-    mqtt_port: int = 0
-    mqtt_username: str = ""
-    mqtt_password: str = ""
-    number_entity_mode: str = "auto"
-    prog_time_interval: int = 15
+    mqtt_host: str = os.getenv("MQTT_HOST", "")
+    mqtt_port: int = int(os.getenv("MQTT_PORT", 0))
```
kellerza, line 57: pylint seems to expect strings as the default arg

```suggestion
    mqtt_port: int = int(os.getenv("MQTT_PORT", "0"))
```

src/ha_addon_sunsynk_multi/options.py
```diff
@@ -52,24 +53,24 @@ class Options:
...
+    mqtt_username: str = os.getenv("MQTT_USERNAME", "")
+    mqtt_password: str = os.getenv("MQTT_PASSWORD", "")
+    number_entity_mode: str = os.getenv("NUMBER_ENTITY_MODE", "auto")
+    prog_time_interval: int = int(os.getenv("PROG_TIME_INTERVAL", 15))
```
kellerza, line 61:
```suggestion
    prog_time_interval: int = int(os.getenv("PROG_TIME_INTERVAL", "15"))
```

kellerza (commented):
If you use env vars, how would you specify the inverters?

hass-addon-sunsynk-multi/README.md
maslyankov: Yes, the idea behind the PR is usage in standalone deployment.

maslyankov:
> If you use env vars, how would you specify the inverters?

Have not looked at this case. The idea behind was to pass MQTT config more easily when using Docker Compose with env vars. I will look into ways to pass inverter config as well but not sure how useful that would be.

maslyankov:
@kellerza Discussed changes have been implemented and functionality verified! Seems to work on my side. I also added documentation on this matter.

src/ha_addon_sunsynk_multi/options.py
```diff
@@ -52,24 +65,30 @@ class Options:
...
+    mqtt_username: str = os.getenv("MQTT_USERNAME", "")
+    mqtt_password: str = os.getenv("MQTT_PASSWORD", "")
+    number_entity_mode: str = os.getenv("NUMBER_ENTITY_MODE", "auto")
+    prog_time_interval: int = int(os.getenv("PROG_TIME_INTERVAL", "15"))
+    inverters: list[InverterOptions] = [
+        InverterOptions(**{k.lower(): v for k, v in inv.items()})
+        for inv in loads(os.getenv("INVERTERS", "[]"))
+    ]
+    sensor_definitions: str = os.getenv("SENSOR_DEFINITIONS", "single-phase")
+    sensors: list[str] = loads(os.getenv("SENSORS", "[]"))
```
kellerza, line 79: We could potentially support simple comma separated strings here? split instead of loads

maslyankov: Yes, but I think this is more consistent with other options. In my opinion it could be added as another way to input a list, but shouldn't replace the json format.

Issue #371, opened 2024-11-24
RW Switches write fail (Bettery Low Capacity, Grid Enable and others maybe)


maslyankov:
Existing sensor definitions can be found [here](https://github.com/kellerza/sunsynk/blob/main/src/sunsynk/definitions/). You can test your own definitions using [custom sensors](https://kellerza.github.io/sunsynk/reference/mysensors)

### Your environment
- Home Assistant version: __
- Addon:
  - Name: **edge**
  - Version: latest
- Inverter:
  - Make: Deye
  - Model: SUN-30K
  - Firmware: 3102-1077-1D08
  - Protocol Version: 0104
- Adaptor details:
  - RS485 via EthernetTCP
  - 
### Describe the issue/bug and what you expect

When you try to set Battery Low Capacity value, it gets set to 0. I have noticed this behavior on other Deye models as well. It can be fixed/set to proper value from Solarman app.

![IMG_0536](https://github.com/user-attachments/assets/749b4a67-8b6e-47bd-8516-fd6a424b7317)

So in conclusion, it works for reading, but not for writing the value of Battery Low Capacity.

This in my opinion is an important config for bad weather and winter months.

Any suggestions on proposed registers is welcome.

### Logs (if applicable)

```logs

paste logs in 3x backticks...

```

gacopl:
Noticed the same here sg04lp3

maslyankov:
Same issue is with "Grid Charge Enabled". When we change setting from Sunsynk mqtt, it goes to an undefined state in Solarman, which results in "False" setting state (Not charging from grid).

maslyankov:
I found what the issue is... So it turns out sunsynk is writing wrong value into modbus registers for some  SwitchRWSensors. They should be 0x0001 (1) for True, but it writes 0x00FF (255) as seen in the logs below: 


```
// Setting to "ON"

sunsynk-mult-amd64-1  | 2024-12-26 21:56:33,152 INFO    sunsynk.sunsynk Writing sensor grid_charge_enabled=ON Registers:{130=0x00ff} 
sunsynk-mult-amd64-1  | 2024-12-26 21:56:33,304 DEBUG   sunsynk.sunsynk Time taken to fetch 1 registers starting at 130 : 0.04s
sunsynk-mult-amd64-1  | 2024-12-26 21:56:33,305 DEBUG   sunsynk.sunsynk Request registers: [130] glen=1. Response [255] len=1. regs={130: 255}
[...]

Issue #374, opened 2024-12-10
SENSORS_FIRST_INVERTER creating same sef of entities on all inverters


gacopl:
### Issue related to

Sunsynk / mbusd Home Assistant Add-On

### Describe the issue/bug

SENSORS_FRIST_INVERTER, impacts just reporting. 
So if I want for instance settings to be available only on Master inverter. 
Addon will produce All inverter devices with same set of entitties, just >1 inverter will show read only entities as unavailable in case of reading sensors. But in case of RW sensors they still work on the >1 inverter. 
I am able to change settings on Slave inverter even though i shouldn't

**Expected behavior**
If i populate SENSORS_FIRST_INVERTER, entities chosen in that section are present only on first inverter and addon produces devices with different entities list

### Your environment

- Home Assistant version: 2024.11.4
- Addon:
  - Name: sunsynk-multi
  - Version: 0.7.5
- Inverter:
  - Make: Deye
  - Model: 12K-SG04LP3
  - Firmware: does not matter
- Adaptor details:
  - RS485 waveshare
  -

### You configuration

```yaml

DRIVER: pymodbus
INVERTERS:
  - SERIAL_NR: "XXX"
    HA_PREFIX: DeyeM
    MODBUS_ID: 1
    DONGLE_SERIAL_NUMBER: "0"
    PORT: tcp://X:502
  - SERIAL_NR: "X"
    HA_PREFIX: DeyeS
    MODBUS_ID: 2
    DONGLE_SERIAL_NUMBER: "0"
    PORT: tcp://X:502
SENSOR_DEFINITIONS: three-phase
SENSORS:
  - energy_management
  - pv2_power
  - voltage
  - load_power
  - power_flow_card
SENSORS_FIRST_INVERTER:
  - settings
MANUFACTURER: Deye
READ_ALLOW_GAP: 6
READ_SENSORS_BATCH_SIZE: 50
SCHEDULES:
  - KEY: W
    READ_EVERY: 5
    REPORT_EVERY: 10
    CHANGE_ANY: true
    CHANGE_BY: 80
    CHANGE_PERCENT: 0
NUMBER_ENTITY_MODE: auto
PROG_TIME_INTERVAL: 15

```

### Logs

```logs

paste logs in 3x backticks...

```

kellerza:
Can you try the latest edge and let me know if this is better?

gacopl:
hmm i dont see any change
```

gacopl:
[...]

PR #377, opened 2024-12-26
Improvement/recursive handle sensor dependencies

maslyankov:
- Fixes https://github.com/kellerza/sunsynk/issues/371 - Grid Charge Enable switch not working
- Handling above issue also helped me find a bug with sensor dependencies being handled only one level down, now handled recursively
- Adds more sensors to groups , updated docs
- Improve docs for custom sensors

## tested and works: 
build: https://github.com/maslyankov/sunsynk/actions/runs/12509593568
image: ghcr.io/maslyankov/hass-addon-sunsynk-multi:edge

## Resulting improvement in config: 

```
SENSORS:
  - energy_management
  - power_flow_card
  - pv_power
  - pv1_power
  - PV1 voltage
  - PV1 current
  - pv2_power
  - PV2 voltage
  - PV2 current
  - Load power
  - Load L1 power
  - Load L2 power
  - Load L3 power
  - Load L1 voltage
  - Load L2 voltage
  - Load L3 voltage
  - battery_temperature
  - Fault
SENSORS_FIRST_INVERTER:
  - settings
  - Grid Charge Battery current
  - Grid Charge Start Battery SOC
  - Grid Charge enabled
  - Battery Max Charge current
  - Battery Max Discharge current
  - Battery Capacity current
  - Date Time
  - Battery Shutdown Capacity
  - Battery Restart Capacity
```

now becomes:

```
SENSORS:
  - energy_management
  - power_flow_card
SENSORS_FIRST_INVERTER:
  - settings
  - diagnostics
```


src/ha_addon_sunsynk_multi/sensor_options.py
```diff
@@ -54,22 +91,12 @@ def init_sensors(self) -> None:
 
         # Add startup sensors
         self.startup = {DEFS.rated_power, DEFS.serial}
-        self[DEFS.rated_power] = SensorOption(
-            sensor=DEFS.rated_power, schedule=Schedule(), visible=False
-        )
-        self[DEFS.serial] = SensorOption(
-            sensor=DEFS.serial, schedule=Schedule(), visible=False
-        )
+        self._add_sensor_with_deps(DEFS.rated_power, visible=False)
```
kellerza, line 94: Is it possible to add rated_power or serial in your config?

kellerza: (or will visible always be false?)

src/ha_addon_sunsynk_multi/sensor_options.py
```diff
@@ -80,18 +107,7 @@ def init_sensors(self) -> None:
...
-            if isinstance(sopt.sensor, RWSensor):
-                for dep in sopt.sensor.dependencies:
-                    self.startup.add(dep)
-                    if dep not in self:
-                        self[dep] = SensorOption(
-                            sensor=dep,
-                            schedule=get_schedule(dep, SCHEDULES),
-                        )
-                    self[dep].affects.add(sopt.sensor)
+                self._add_sensor_with_deps(sen, visible=True)
```
kellerza, line 110: Don't we need a first=True here as well? (passed to the SensorOptions?) - #374 made me think of this

src/ha_addon_sunsynk_multi/sensor_options.py
maslyankov: I made it so that all dependencies that are not defined in the config are hidden by default. There are further improvements to this branch on other branches, but my repo has diverged from this one somewhat.

src/ha_addon_sunsynk_multi/sensor_options.py
maslyankov: I made this change because of failing tests. Will look into it if it is necessary.

kellerza:
Thanks @maslyankov - this seems to fix some startup (restart) issues for me as well

Issue #390, opened 2025-01-15
Deye/Sunsync 3 phase check "Total Active Power from Side to Side of the Grid"


proggaras:
Existing sensor definitions can be found [here](https://github.com/kellerza/sunsynk/blob/main/src/sunsynk/definitions/). You can test your own definitions using [custom sensors](https://kellerza.github.io/sunsynk/reference/mysensors)

### Your environment

- Home Assistant version: 2025.1.0
- Addon:
  - Name: edge/dev
  - Version: 0.8.2
- Inverter:
  - Make: Deye
  - Model: SUN-25K-SG01HP3-EU-AM2
  - Firmware: HMI: Ver 2001-C030 MAIN: Ver 3002-1066-1E08
- Adaptor details:
  - RS485
  - Waveshare RS485_TO_POE_ETH

### Describe the issue/bug and what you expect

With this [commit](https://github.com/kellerza/sunsynk/commit/e6a5c0b6bb343ec01d69be11eac33a62fa75f117) the value 607, "Total Active Power from Side to Side of the Grid" was changed from unsigned to signed. 

According to the [Documentation](https://forum.iobroker.net/assets/uploads/files/1712644425833-c3ca14a0-178c-4f68-be8c-fa68a414b97b-deye-sun-inverter-modbus-en.pdf) 607 has no S16 flag in the last column.
Instead 604, 605, 606 has the S16 flag. 

Is that correct? Please let me know if I should create a pull request. 

<img width="588" alt="image" src="https://github.com/user-attachments/assets/a8f04409-2b13-4c68-8085-2775906402cb" />

maslyankov:
Probably you are looking at an old version of the documentation. Here you can see (the most up to date one I could find) that this register also has S16 notation: https://github.com/maslyankov/sunsynk/blob/main/inverter-docs/deye/three-phase/MODBUS%20RTU%E4%B8%89%E7%9B%B8%E5%82%A8%E8%83%BD%E9%80%9A%E4%BF%A1%E8%A7%84%E7%BA%A6V105.3-20240710-Lee_Dobson.docx

kellerza:
If the individual phases are signed, it makes sense that the total is signed.  PR welcome to update 604-606 to signed.

The names seem to be a direct translation, but I don't have a suggestion to improve them at the moment. (If you do change them the old/new need to be added to the deprecated dict as well)

maslyankov:
[...]

Issue #381, opened 2025-01-02
Wrong "ON" value for "Solar Export"


rrveex:
### Your environment
- Inverter:
  - Make: Deye
  - Model: SUN-15K-SG05LP3-EU-SM2
  - Firmware: "Ver 2021-1147-1807"

### Describe the issue/bug and what you expect

![image](https://github.com/user-attachments/assets/235defed-387a-4be6-8f08-4de1bc555ba4)

If I toggle this to "on", value `255` is written to modbus address `145`. The inverter *doesn't* turn on export.

If I (manually) write the value `1` to address `145`, the inverter starts exporting.

Value `0` (either manually or via GUI toggle) correctly stops export.

maslyankov:
I have fixed it. Please try the following docker image: ghcr.io/maslyankov/hass-addon-sunsynk-multi:edge
Please let me know if it works or not. Thanks!

rrveex:
Thank you for fixing!

Sorry, I'm running HA in a VM, it'd take probably a day or so for me to learn how to run and (re-)configure it in docker for testing. A quick search on how I could get to the .py file directly didn't help.

BTW, `112` (activate battery) - the doc is wrong, 0=disabled, 1=enabled, tested by checking/unchecking in the HMI and querying modbus.

maslyankov:
> Thank you for fixing!
> 
> Sorry, I'm running HA in a VM, it'd take probably a day or so for me to learn how to run and (re-)configure it in docker for testing. A quick search on how I could get to the .py file directly didn't help.
> 
> BTW, `112` (activate battery) - the doc is wrong, 0=disabled, 1=enabled, tested by checking/unchecking in the HMI and querying modbus.

Thank you for verifying this! Will fix this as well. Meanwhile, you do not need to change your usage approach. You can use this as addon: https://sunsynk.wectrl.net/guide/getting-started

rrveex:
![image](https://github.com/user-attachments/assets/7f022b70-5027-4df6-b167-89fd05f0e7e8)

Works!

kellerza:
@rrveex this is not yet fixed if in a different fork

Two solutions could be on=1 or mask=1 (@rixxxx do you use this setting?)

@maslyankov - is the intention to create a PR/maintain your own fork?

maslyankov:
> @rrveex this is not yet fixed if in a different fork
> 
> Two solutions could be on=1 or mask=1 (@rixxxx do you use this setting?)
> 
[...]

PR #391, opened 2025-01-15
Upstream/general improvements

maslyankov:
Improvements to sensor definitions for single and three-phase models. Ass well as new / updated sensor groups.

Sorry for big PR...

Addresses: 
- https://github.com/kellerza/sunsynk/issues/390
- https://github.com/kellerza/sunsynk/issues/381#issuecomment-2568995917


src/sunsynk/definitions/three_phase_common.py
```diff
@@ -291,7 +298,16 @@
 # System program
 #################
 
-SENSORS += SwitchRWSensor(141, "Priority Load")
+SENSORS += (
```
kellerza, line 301: on the power flow card this is a simple switch. On=load first

So lets keep it the same as the single phase definition & aligned with the power flow card

src/sunsynk/definitions/three_phase_common.py
```diff
@@ -532,6 +548,14 @@
             0b11 << 14: "Enable",
         },
         bitmask=0b11 << 14,
+    ),
+        SelectRWSensor(
```
kellerza, line 552: This can also be a simple SwitchRWSensor

src/sunsynk/sensors.py
```diff
@@ -267,7 +267,7 @@ def reg_to_value(self, regs: RegType) -> ValType:
             for bit in range(16):
                 msk = 1 << bit
                 if msk & b16:
-                    msg = f"F{bit + off + 1:02} " + faults.get(off + msk, "")
+                    msg = f"F{bit+off+1:02} " + faults.get(off + msk, "")
```
kellerza, line 270: Can you please format with ruff/ use the pre-commit hooks? (I believe that will add back these spaces)

inverter-docs/deye/single-phase/Modbus.-.-.V118.pdf
kellerza, line 1: I'm not sure we have permission to distribute these protocol documents?

Some even have personal info in the filename

kellerza (changes requested):

kellerza:
> @kellerza Please review my handling of the issues with recursive sensor dependencies here: https://github.com/maslyankov/sunsynk/blob/main/src/ha_addon_sunsynk_multi/sensor_options.py (https://github.com/maslyankov/sunsynk/blob/39262f08ae6bd89dfefd123531e9eaebef435e48/src/ha_addon_sunsynk_multi/sensor_options.py#L49C1-L50C1)
> 
> I have not added it to this PR. I'd like to hear your thoughts on it.

The biggest drawback with the recursive method is that checking if a sensor (or its deps) is visible/first only is quite complex.

With the current implementation [sensor_options](https://github.com/kellerza/sunsynk/blob/main/src/ha_addon_sunsynk_multi/sensor_options.py#L78) they are added in the following order, and visibility/first seems depends on the order. Adding deps are also delayed (since they could be visible)

The order, as comments in the code
1. Add startup sensors to all inverters. Visible if configured anywhere.
2. Add sensors configured for all inverters
3. Add deps, usually hidden, but visible if configured on 1st inverter
4. Add sensors configured for the 1st inverter
5. Add deps for the 1st inverter, always hidden

maslyankov:
> > @kellerza Please review my handling of the issues with recursive sensor dependencies here: https://github.com/maslyankov/sunsynk/blob/main/src/ha_addon_sunsynk_multi/sensor_options.py (https://github.com/maslyankov/sunsynk/blob/39262f08ae6bd89dfefd123531e9eaebef435e48/src/ha_addon_sunsynk_multi/sensor_options.py#L49C1-L50C1)
> > I have not added it to this PR. I'd like to hear your thoughts on it.
> 
> The biggest drawback with the recursive method is that checking if a sensor (or its deps) is visible/first only is quite complex.
> 
> With the current implementation [sensor_options](https://github.com/kellerza/sunsynk/blob/main/src/ha_addon_sunsynk_multi/sensor_options.py#L78) they are added in the following order, and visibility/first seems depends on the order. Adding deps are also delayed (since they could be visible)
> 
> The order, as comments in the code
> 
> 1. Add startup sensors to all inverters. Visible if configured anywhere.
> 2. Add sensors configured for all inverters
> 3. Add deps, usually hidden, but visible if configured on 1st inverter
> 4. Add sensors configured for the 1st inverter
> 5. Add deps for the 1st inverter, always hidden

In the current implementation, do we add sensors as visible if they are defined in sensor group?

kellerza:
> In the current implementation, do we add sensors as visible if they are defined in sensor group?

Yes, the groups are expanded right at the start. Nothing depends on a group

maslyankov:
> Owner

Okay. Any other changes to add to this branch?

src/sunsynk/definitions/three_phase_common.py
kellerza: It seems you changed the wrong one?

141 & 142 should probably be unchanged

maslyankov: Yes, sorry, my mistake. Should be okay now.

kellerza (approved):

kellerza:
Thanks @maslyankov !

Doesi:
Since these changes, power sensors on my DEYE SUN 20K-SG01 HP3-EU-AM2 randomly jump to 65535 kW and -65509 kW. This has never happened before.

proggaras:
I can confirm. Here is an example for "Inverter L1 Power": 
image

maslyankov:
Can you create an issue? I think I have an idea why this might happen and a PR for possible fix. Thanks.

PR #393, opened 2025-01-16
Deye 3phase power switch renamed to Inverter enabled

proggaras:
New name changed from Switch "Inverter Power" to "Inverter enabled" to avoid override.


kellerza:
Thx, can you include the single phase definition?

PR #397, opened 2025-01-18
Enhance helper functions for register value conversions

maslyankov:
- Added `make_32bit` to convert two 16-bit values into a 32-bit integer.
- Introduced `split_to_16bit` to split a 32-bit integer into two 16-bit values.
- Updated `signed` function to handle struct packing for signed integers.
- Modified `NumberRWSensor` to utilize `split_to_16bit` for handling two 16-bit registers.
- Refactored `reg_to_value` in `Sensor` class to use `make_32bit` for combining register values.

Related PR & issues that arised from it:
- https://github.com/kellerza/sunsynk/pull/391#issuecomment-2599665488
- https://github.com/kellerza/sunsynk/pull/391#issuecomment-2599698838
- https://github.com/kellerza/sunsynk/issues/398

## Why?

Using struct for handling signed integers (especially when merging registers) is better for several key reasons:

1. **Proper Two's Complement Handling**:
   - Struct automatically handles two's complement representation for signed numbers
   - Direct bit manipulation might miss the sign bit handling, leading to incorrect negative values

2. **Platform Independence**:
   ```python
   # Using struct (safe)
   import struct
   value = struct.unpack('>i', bytes([0xFF, 0xFF, 0x80, 0x00]))[0]  # Correctly handles -32768
   
   # Direct bit manipulation (risky)
   value = (register1 << 16) | register2  # Might not handle negative numbers correctly
   ```

3. **Endianness Handling**:
   - Struct allows explicit specification of byte order (big-endian '>' or little-endian '<')
   - Direct bit manipulation requires manual handling of endianness

4. **Type Safety**:
   - Struct enforces proper data type conversion
   - Helps prevent overflow issues and undefined behavior

5. **Maintainability**:
   - Code intent is clearer with struct
   - Less prone to bugs when dealing with different register sizes or signed/unsigned conversions

6. **Standard Library Solution**:
   - Well-tested and documented
   - Consistent behavior across Python versions and platforms

The small performance overhead of using struct is usually worth the reliability and clarity it provides when handling signed integers.

### DIFF
--- a/src/sunsynk/helpers.py
+++ b/src/sunsynk/helpers.py
@@ -26,9 +26,15 @@ def pack_value(value: int, bits: int = 16, signed: bool = True) -> RegType:
         For 32-bit: tuple of (low, high) register values
     """
     if bits == 16:
+        if not signed and value < 0:
+            # Convert negative value to two's complement for unsigned
+            value = value + (1 << 16)
         fmt = 'h' if signed else 'H'
         return struct.unpack('H', struct.pack(fmt, value))
     if bits == 32:
+        if not signed and value < 0:
+            # Convert negative value to two's complement for unsigned
+            value = value + (1 << 32)
         fmt = 'i' if signed else 'I'
         return struct.unpack('2H', struct.pack(fmt, value))
     raise ValueError(f"Unsupported number of bits: {bits}")
--- a/src/sunsynk/helpers.py
+++ b/src/sunsynk/helpers.py
@@ -13,7 +13,7 @@ RegType = tuple[int, ...]
 NumType = float | int
 
 
-def pack_value(value: int, bits: int = 16, signed: bool = True) -> int | tuple[int, int]:
+def pack_value(value: int, bits: int = 16, signed: bool = True) -> RegType:
     """Pack a value into register format.
     
     Args:
@@ -27,7 +27,7 @@ def pack_value(value: int, bits: int = 16, signed: bool = True) -> int | tuple[i
     """
     if bits == 16:
         fmt = 'h' if signed else 'H'
-        return struct.unpack('H', struct.pack(fmt, value))[0]
+        return struct.unpack('H', struct.pack(fmt, value))
     if bits == 32:
         fmt = 'i' if signed else 'I'
         return struct.unpack('2H', struct.pack(fmt, value))
@@ -102,6 +102,11 @@ def hex_str(regs: RegType, address: RegType | None = None) -> str:
     return f"{{{' '.join(res)}}}"
 
 
+def patch_bitmask(value: int, patch: int, bitmask: int) -> int:
+    """Combine bitmask values."""
+    return (patch & bitmask) + (value & (0xFFFF - bitmask))
+
+
 class SSTime:
     """Deals with inverter time format conversion complexities."""
 
@@ -151,8 +156,3 @@ class SSTime:
             self.minutes = int(hours) * 60 + int(minutes)
         except ValueError:
             _LOGGER.warning("Invalid time string: %s (expected hh:mm)", value)
-
-
-def patch_bitmask(value: int, patch: int, bitmask: int) -> int:
-    """Combine bitmask values."""
-    return (patch & bitmask) + (value & (0xFFFF - bitmask))
--- a/src/sunsynk/helpers.py
+++ b/src/sunsynk/helpers.py
@@ -26,15 +26,9 @@ def pack_value(value: int, bits: int = 16, signed: bool = True) -> RegType:
         For 32-bit: tuple of (low, high) register values
     """
     if bits == 16:
-        if not signed and value < 0:
-            # Convert negative value to two's complement for unsigned
-            value = value + (1 << 16)
         fmt = 'h' if signed else 'H'
         return struct.unpack('H', struct.pack(fmt, value))
     if bits == 32:
-        if not signed and value < 0:
-            # Convert negative value to two's complement for unsigned
-            value = value + (1 << 32)
         fmt = 'i' if signed else 'I'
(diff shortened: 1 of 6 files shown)


maslyankov:
I will work on tests and linting, but first @kellerza I'd like to hear your thoughts on this.

JanKraslice:
How do I get back to the 2024 version? I'm keeping it permanently. We can't play here.
Accuracy has to be 99.9%. Anything less is wrong and we can't deploy it.
Thank you kellerza I trust you

maslyankov:
> How do I get back to the 2024 version? I'm keeping it permanently. We can't play here.
> 
> Accuracy has to be 99.9%. Anything less is wrong and we can't deploy it.
> 
> Thank you kellerza I trust you

Bro stop spamming around like that. This is an "edge" version - still in testing. Use the release version if it's that serious of an issue for you. Everyone here is working free of charge to make the project better.

src/sunsynk/rwsensors.py
```diff
@@ -78,7 +78,8 @@ def value_to_reg(self, value: ValType, resolve: ResolveType) -> RegType:
                 val = 0x10000 + val
```
kellerza, line 78: This one still uses an unsafe bitwise manipulation

src/sunsynk/sensors.py
```diff
@@ -38,11 +39,14 @@ def id(self) -> str:
...
-        if len(regs) > 1:
-            val += regs[1] << 16
-        if self.factor < 0:  # Indicates this register is signed
-            val = signed(val, bits=16 * len(regs))
+        if len(regs) == 1:
+            val = regs[0]
+            if self.factor < 0:  # Indicates this register is signed
+                val = signed(val, bits=16)
+        elif len(regs) == 2:
+            val = make_32bit(regs[0], regs[1], signed=self.factor < 0)
```
kellerza, line 47: Can we combine signed & make_32bit?

```python
def unpack(*regs: int, *, signed: bool = True, bits: int=16) -> int:
```

kellerza (commented):
No objections to using std lib struct

It does feel like it can be a bit more unified. I made a proposal.
As long as we have tests then it should be ok.

src/sunsynk/sensors.py
kellerza: This iwll break on of the signed tests of today, but in those test, 32-bit must be split into 2x regs

kellerza:
> How do I get back to the 2024 version? I'm keeping it permanently. We can't play here. Accuracy has to be 99.9%. Anything less is wrong and we can't deploy it. Thank you kellerza I trust you

Like @maslyankov - either use the multi version, or downgrade using the HA supervisor / point to another docker image

maslyankov:
I have streamlined the implementation. @kellerza please take a look if I've missed anything. I will then work on passing linting and passing/updating tests.

src/sunsynk/helpers.py
```diff
@@ -12,6 +13,46 @@
 NumType = float | int
 
 
+def pack_value(value: int, bits: int = 16, signed: bool = True) -> int | tuple[int, int]:
```
kellerza, line 16: -> RegType

src/sunsynk/helpers.py
```diff
@@ -48,21 +89,19 @@ def as_num(val: ValType) -> float | int:
...
-    # 16-bit only (old)
-    # return val if val <= 0x7FFF else val - 0x10000
-
-
 def slug(name: str) -> str:
     """Create a slug."""
     return name.lower().replace(" ", "_").replace("-", "_")
 
 
+def hex_str(regs: RegType, address: RegType | None = None) -> str:
```
kellerza, line 97: What changed?

src/sunsynk/rwsensors.py
```diff
@@ -68,17 +76,17 @@ def dependencies(self) -> list[Sensor]:
...
         if len(self.address) == 1:
-            if val < 0:
-                val = 0x10000 + val
-            return self.reg(val)
+            return self.reg(pack_value(val, bits=16, signed=self.factor < 0))
         if len(self.address) == 2:
-            return self.reg(val & 0xFFFF, int(val >> 16))
+            low, high = pack_value(val, bits=32, signed=self.factor < 0)  # type: ignore
+            return self.reg(low, high)
         raise NotImplementedError(f"Address length not supported: {self.address}")
```
kellerza, line 90:
```suggestion
return self.reg(*pack_value(val, bits=len(self.addrees)*16, signed=self.factor < 0))
```

src/sunsynk/helpers.py
```diff
@@ -12,6 +13,46 @@
...
+        bits: Number of bits (16 or 32)
+        signed: Whether the value should be treated as signed
+    
+    Returns:
+        For 16-bit: single register value
+        For 32-bit: tuple of (low, high) register values
+    """
+    if bits == 16:
+        fmt = 'h' if signed else 'H'
+        return struct.unpack('H', struct.pack(fmt, value))[0]
```
kellerza, line 30:
```suggestion
        return struct.unpack('H', struct.pack(fmt, value))
```

src/tests/sunsynk/test_helpers.py
```diff
@@ -96,7 +105,23 @@ def test_time() -> None:
     assert time.minutes == just_before_midnight
 
 
+def test_pack_unpack() -> None:
+    """Test pack_value and unpack_value functions."""
+    # Test 16-bit values
```
kellerza, line 110: In test_rwsensors, the sensor is not signed (that test case needs fixing factor to -1)

Can we add a test here to show we expect it to fail?

```python
wiih pytest.raises(struct.error):
  pack_value(-1, bits=16, signed=False)
```

PR #399, opened 2025-01-20
Update single_phase.py

deondti:
change max charge/discharge current to 290A to accommodate the 16KW Deye single phase 

I'd much rather propose ( not sure if possible)
Have a different inverter selection in configuration or;
if sensor.rated power=x then set charge amps to Y
else if sensor.rated power= a then set charge amps to B etc.

### DIFF
--- a/src/sunsynk/definitions/single_phase.py
+++ b/src/sunsynk/definitions/single_phase.py
@@ -311,9 +311,9 @@ SENSORS += (
 #############
 # Inverter settings
 #############
-SENSORS += NumberRWSensor(230, "Grid Charge Battery current", AMPS, min=0, max=185)
-SENSORS += NumberRWSensor(210, "Battery Max Charge current", AMPS, min=0, max=185)
-SENSORS += NumberRWSensor(211, "Battery Max Discharge current", AMPS, min=0, max=185)
+SENSORS += NumberRWSensor(230, "Grid Charge Battery current", AMPS, min=0, max=290)
+SENSORS += NumberRWSensor(210, "Battery Max Charge current", AMPS, min=0, max=290)
+SENSORS += NumberRWSensor(211, "Battery Max Discharge current", AMPS, min=0, max=290)
 
 
 #############
--- a/src/sunsynk/definitions/single_phase.py
+++ b/src/sunsynk/definitions/single_phase.py
@@ -311,7 +311,7 @@ SENSORS += (
 #############
 # Inverter settings
 #############
-SENSORS += NumberRWSensor(230, "Grid Charge Battery current", AMPS, min=0, max=290)
+SENSORS += NumberRWSensor(230, "Grid Charge Battery current", AMPS, min=0, max=140)
 SENSORS += NumberRWSensor(210, "Battery Max Charge current", AMPS, min=0, max=290)
 SENSORS += NumberRWSensor(211, "Battery Max Discharge current", AMPS, min=0, max=290)
 


maslyankov:
This could introduce issues with smaller inverters. Maybe we can integrate a variable for Inverter power rating, or get it dynamically from the inverter model register (I haven't verified if such is available) and assign max values for some things based on that.

deondti:
@maslyankov  I agree my solution is not the best plan ,  there is sensor.rated_power
I'm just not sure how to implement a dynamic solution, global options based on inverter rating, 
that's why my other suggestion is: more inverter selections in configuration,  if that solution is easier.

I'd hate to mess up this awesome project of @kellerza

Waz-Cpt:
It should be relatvely easy to gather the list singles phase power ratings and then define their respective amp rating based on the operational voltage. This could be a theoretical calculated value give operational voltage and rated power values.

I am not sure how models for 120V and 230V volts would differ on this basis, but there definitely would be a difference.

The other challenge is that these inverters are designed with a rated power, but can sustain overload situations for brief periods of time.

It may be interesting if someone in the community does the leg work to document all the single phase and three phase inverters rated power, peak overload power, operation voltage and charge/discharge limts.

Developing a solution of programmatically assigning the values would need the data to ensure safe operation is maintained.

deondti:
> It should be relatvely easy to gather the list singles phase power ratings and then define their respective amp rating based on the operational voltage. This could be a theoretical calculated value give operational voltage and rated power values.
> 
> I am not sure how models for 120V and 230V volts would differ on this basis, but there definitely would be a difference.
> 
> The other challenge is that these inverters are designed with a rated power, but can sustain overload situations for brief periods of time.
> 
> It may be interesting if someone in the community does the leg work to document all the single phase and three phase inverters rated power, peak overload power, operation voltage and charge/discharge limts.
> 
> Developing a solution of programmatically assigning the values would need the data to ensure safe operation is maintained.

I hear you about the peak values(overload) but we need to maintain the maximum values allowed that one can set on the display else we run into other problems. We cannot use theoretical values.
eg. The 16KW only allows 290A from the battery which does not allow for the full 16KW on battery mode when the sun is down,

maslyankov:
> @maslyankov I agree my solution is not the best plan , there is sensor.rated_power I'm just not sure how to implement a dynamic solution, global options based on inverter rating, that's why my other suggestion is: more inverter selections in configuration, if that solution is easier.
> 
> I'd hate to mess up this awesome project of @kellerza
> 
> There is also the issue of the 16KW inverter only reading on grid eg. 500W where it is actually 5000W, custom sensors solve the issue, no hassles Sensor(172, "Grid_Ct_Power_10", WATT, -10), Sensor(169, "Grid_power_10", WATT, -10),
> 
> This could also be solved in the same manner

For the grid power it's strange as in docs it is noted as:

image




> Are you aware that the 16kW is a special case. As you appear to be aware, the 16kW is not actually 16kW, it is actually 13kW inverter.
> 
> Also the spec sheets indicate that the discharge and charge on the 16kw are not the same, 275A for max charge and 300A max discharge.
> 
> It would be quite the challenge to document all the various inverters and have all their max and min values stored within the integration.
> 
> If these max and min values existed as registers than it would be great.

There are sadly no such registers, that I can find.

Waz-Cpt:
@maslyankov I thought these max or min values would not be in any registers. A set of definitions could be setup to assist in assigning max and min values for some sensors, but it is a challenge to have this data maintained.

I think it could be included as a option in the addon config to define the max and min values for this and possibly other sensors.

But decent disclaimer waving blame on the devs is crucial for users inputing values that could damage their supply, batteries or inverters.

deondti:
> Are you aware that the 16kW is a special case. As you appear to be aware, the 16kW is not actually 16kW, it is actually 13kW inverter.
> 
> Also the spec sheets indicate that the discharge and charge on the 16kw are not the same, 275A for max charge and 300A max discharge.
> 
> It would be quite the challenge to document all the various inverters and have all their max and min values stored within the integration.
> 
> If these max and min values existed as registers than it would be great.

Hi, I do realize that the 16Kw is unique if not "a problem"
Interesting figures quoted, on my DEYE 16KW SA version Both charge and discharge allow for 290A max as it came out of the factory, perhaps we are sitting with different versions of the 16KW aswell

Waz-Cpt:
From three of my installer friends and one of my friends that is a reseller of Deye, the numbers on most of the 16kW datasheets is not actually valid and is misleading. It is the one big problem child out of all the inverters from Sunsynk and Deye.

It technically can only push around 13kW from the Batteries, however can handly a very high grid passthrough. There are some major limitations with it that are not advertised. So much so that two of the installers I know refuse to install them.

src/sunsynk/definitions/single_phase.py
```diff
@@ -311,9 +311,9 @@
 #############
 # Inverter settings
 #############
-SENSORS += NumberRWSensor(230, "Grid Charge Battery current", AMPS, min=0, max=185)
-SENSORS += NumberRWSensor(210, "Battery Max Charge current", AMPS, min=0, max=185)
-SENSORS += NumberRWSensor(211, "Battery Max Discharge current", AMPS, min=0, max=185)
+SENSORS += NumberRWSensor(230, "Grid Charge Battery current", AMPS, min=0, max=290)
```
deondti, line 314: Grid Charge should be max 140A
The rest is correct as per lcd menu

kellerza:
@deondti - I think the correct approach would be to introduce a new variant of today's single-phase definition

Today we have these definitions:
```yaml
SENSOR_DEFINITIONS: list(single-phase|three-phase|three-phase-hv)
```

We could use the following options, which will enable us to dynamically import the correct definition file, instead of today's explicit imports

```yaml
SENSOR_DEFINITIONS: list(single-phase|single-phase-16kw|three-phase-lv|three-phase-hv)
```

For this PR, I'm ok if you create a `single_phase_16kw.py`, then extend the sensors from single_phase, similar to the way we extend the 3-phase definitions. This allows you to fix the max & all the factors

I can take care of the import logic later, except if you are up for it?

deondti:
> @deondti - I think the correct approach would be to introduce a new variant of today's single-phase definition
> 
> Today we have these definitions:
> 
> ```yaml
> SENSOR_DEFINITIONS: list(single-phase|three-phase|three-phase-hv)
> ```
> 
> We could use the following options, which will enable us to dynamically import the correct definition file, instead of today's explicit imports
> 
> ```yaml
> SENSOR_DEFINITIONS: list(single-phase|single-phase-16kw|three-phase-lv|three-phase-hv)
> ```
> 
> For this PR, I'm ok if you create a `single_phase_16kw.py`, then extend the sensors from single_phase, similar to the way we extend the 3-phase definitions. This allows you to fix the max & all the factors
> 
> I can take care of the import logic later, except if you are up for it?

I Think at this stage it would be the best solution, lets do it

deondti:
@kellerza 
It's done :)

Thank you once again for this great project and being able to contribute

kellerza:
#405

PR #405, opened 2025-01-22
Create single_phase_16kw.py

deondti:
Creating a " single_phase_16kw.py" definition.
Major changes from 5/8Kw to 16Kw
    Sensor(172, "Grid_Ct_Power_10", WATT, -10),
    Sensor(169, "Grid_power_10", WATT, -10),

SENSORS += NumberRWSensor(230, "Grid Charge Battery current", AMPS, min=0, max=140) SENSORS += NumberRWSensor(210, "Battery Max Charge current", AMPS, min=0, max=290) SENSORS += NumberRWSensor(211, "Battery Max Discharge current", AMPS, min=0, max=290)

Battery charge MAX is derived from the physical inverter LCD settings.

Grid and CT power is because it would only read eg. 500W instead of 5000W

### DIFF
--- a/src/sunsynk/definitions/single_phase_16kw.py
+++ b/src/sunsynk/definitions/single_phase_16kw.py
@@ -1,0 +1,351 @@
+"""Sunsynk / Deye 16kW hybrid inverter sensor definitions."""
+
+from sunsynk import AMPS, CELSIUS, KWH, VOLT, WATT
+from sunsynk.definitions import COMMON, PROG_CHARGE_OPTIONS, PROG_MODE_OPTIONS
+from sunsynk.rwsensors import (
+    NumberRWSensor,
+    SelectRWSensor,
+    SwitchRWSensor,
+    SystemTimeRWSensor,
+    TimeRWSensor,
+)
+from sunsynk.sensors import (
+    BinarySensor,
+    FaultSensor,
+    InverterStateSensor,
+    MathSensor,
+    SDStatusSensor,
+    Sensor,
+    TempSensor,
+)
+
+SENSORS = COMMON.copy()
+
+##########
+# Battery
+##########
+SENSORS += (
+    TempSensor(182, "Battery temperature", CELSIUS, 0.1),
+    Sensor(183, "Battery voltage", VOLT, 0.01),
+    Sensor(184, "Battery SOC", "%"),
+    Sensor(190, "Battery power", WATT, -1),
+    Sensor(191, "Battery current", AMPS, -0.01),
+    # Charge and Discharge limit vary based on temperature and SoC
+    Sensor(314, "Battery Charge Limit current", AMPS, -1),  # #327
+    Sensor(315, "Battery Discharge Limit current", AMPS, -1),  # #327
+)
+
+###########
+# Inverter
+###########
+SENSORS += (
+    Sensor(175, "Inverter power", WATT, -1),
+    Sensor(154, "Inverter voltage", VOLT, 0.1),
+    Sensor(164, "Inverter current", AMPS, 0.01),
+    Sensor(193, "Inverter frequency", "Hz", 0.01),
+)
+
+#######
+# Grid
+#######
+SENSORS += (
+    Sensor(79, "Grid frequency", "Hz", 0.01),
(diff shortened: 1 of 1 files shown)


kellerza:
If you import from the single_phase you only need to define changes & not all sensors

look at 3ph lv & hv to see how this is done
Import SENSORS, create a copy and add sensors to overwrote ones with the same name

deondti (commented):
corrected?

src/sunsynk/definitions/single_phase_16kw.py
```diff
@@ -0,0 +1,40 @@
+"""Sunsynk / Deye 16kW hybrid inverter sensor definitions."""
+
+from sunsynk import AMPS, CELSIUS, KWH, VOLT, WATT
+from sunsynk.definitions import COMMON, PROG_CHARGE_OPTIONS, PROG_MODE_OPTIONS
```
kellerza, line 4: You need Ll the single phase ones as base

```suggestion
from sunsynk.definitions.single_phase import SENSORS as COMMON
```

kellerza:
The selection mechanism is still pending

kellerza:
Just commit changes to this branch, then it will update

deondti:
> Just commit changes to this branch, then it will update

Please help , I think I'm stuck.
To me it looks like the file is in the branch but I don't think it is...
or am I just being a noob ?

kellerza:
The best would be to install pre-commit locally & use that to fix up your tests that are failing.

```bash
pip install pre-commit
git add . && pre-commit run --all
```

I implemented dynamic loading, so you should just add single-phase-16kw to config.yaml in the edge addon folder

you can also rebase on the latest changes (after a commit!) using `git pull --rebase`, maybe you need to include "upstream main" as part of your rebase

src/sunsynk/definitions/single_phase_16Kw.py
```diff
@@ -0,0 +1,40 @@
...
+)
+   
+#############
+# Inverter settings
+#############
+SENSORS += NumberRWSensor(230, "Grid Charge Battery current", AMPS, min=0, max=140)
+SENSORS += NumberRWSensor(210, "Battery Max Charge current", AMPS, min=0, max=290)
+SENSORS += NumberRWSensor(211, "Battery Max Discharge current", AMPS, min=0, max=290)
+
+)
```
kellerza, line 40: You probably need to delete line 40 to make the test pass?

deondti:
Hi @kellerza 

Am I on the right track?

Could you please help with these last 2 errors? I have no idea what it means...

"" 
************* Module sunsynk.definitions.single_phase_16kw
src/sunsynk/definitions/single_phase_16kw.py:32:0: C0303: Trailing whitespace (trailing-whitespace)
src/sunsynk/definitions/single_phase_16kw.py:40:0: C0305: Trailing newlines (trailing-newlines)
""

hass-addon-sunsynk-edge/config.yaml
```diff
@@ -59,7 +59,7 @@ schema:
       MODBUS_ID: int(1,16)
       DONGLE_SERIAL_NUMBER: str?
       PORT: str
-  SENSOR_DEFINITIONS: list(single-phase|three-phase|three-phase-hv)
+  SENSOR_DEFINITIONS: list(single-phase|single-phase-16Kw|three-phase|three-phase-hv)
```
kellerza, line 62: Either all names needs to be lowercase, or we need a `.lower()` in the load function in definitions/__init__.py

Kw will not match the filename kw today

kellerza:
Thanks @deondti !

you should be able to test this once the edge version is built

hass-addon-sunsynk-edge/config.yaml
kellerza: Codecov should be taken care of by a previous commit. It imports and tests all the existing definition files.

deondti:
sorry for the lengthy mess and THANK YOU for your patience

deondti:
Tested and seems to be fine, thank you again for this great project

talabirast:
I haven't fully tested it yet, but it seems that this mode also works on  deye single_phase_18kw

PR #414, opened 2025-01-24
Update multi-options.md

deondti:
added sensor single-phase-16Kw


kellerza:
You can add this change to #405

PR #460, opened 2025-04-10
Max Solar power limits and HV Grid Peak Shaving

rixxxx:
Increase Max Solar power upper limits and set minimum for 3PH HV. Limit below 1000 doesn't do anything.
Bring back Grid Peak Shaving for 3PH HV.


martinarva:
Thanks @rixxxx for the PR! 🙌
@kellerza, could you please merge it when you get a chance? I’m currently unable to use Max Solar Power in one of my automations, as the cap is stuck at 12,000 and I need to set it back to 15,600.

PR #485, opened 2025-06-19
restore correct scale for 3ph grid peak shaving

rixxxx:


rixxxx:
> duplicate of #486?

I planned to separate the things to different PRs, but probably failed to do so. #486 still included #485

Issue #510, opened 2025-08-14
Sunsynk 7kW Hybrid + USR-W630


hamid-elaosta:
### Issue related to

Sunsynk / mbusd Home Assistant Add-On

### Describe the issue/bug

Lots of errors parsing and reading from the MODBUS device.

I'm new to this so I'm not sure if the inverter isn't supported, or if the W630 is misconfigured or the wire is too long for the RS485 (~3M) or something other config is wrong. Any help appreciated.

**Expected behavior**
Connects and reads values from inverter

### Your environment

- Home Assistant version: 2025.8.1
- Addon:
  - Name: sunsynk-multi
  - Version: 0.8.2
- Inverter:
  - Make: Sunsynk
  - Model: SYNK-7K-SG05LP1
  - Firmware: M 5.3.8.5 / S 1.5.1.5 / C E.4.3.8
- Adaptor details:
  - RS485 on USR-W630
  - Purchased CAN/RS485 Splitter connected to dual BMS CAN/RA485 port.

### You configuration

All sensor values as default when add-on installed, only serial number, port and batch size changed.

```yaml

DRIVER: pymodbus
INVERTERS:
  - SERIAL_NR: "XXXXXXXX"
    HA_PREFIX: sunsynk
    MODBUS_ID: 1
    DONGLE_SERIAL_NUMBER: "0"
    PORT: tcp://10.0.66.129:8899
SENSOR_DEFINITIONS: single-phase
SENSORS:
  - energy_management
  - power_flow_card
  - pv2_power
SENSORS_FIRST_INVERTER:
  - settings
MANUFACTURER: Sunsynk
READ_ALLOW_GAP: 2
READ_SENSORS_BATCH_SIZE: 8
SCHEDULES:
  - KEY: W
    READ_EVERY: 5
    REPORT_EVERY: 60
    CHANGE_ANY: false
    CHANGE_BY: 80
    CHANGE_PERCENT: 0
NUMBER_ENTITY_MODE: auto
PROG_TIME_INTERVAL: 15
MQTT_HOST: core-mosquitto
MQTT_PORT: 1883
MQTT_USERNAME: hass
MQTT_PASSWORD: ""


```

### Logs

```logs

############################################################
12:02:46 INFO                 Inverter serial number '****XXXX'
12:02:46 INFO    ############################################################
12:02:46 INFO    Device type: Single phase hybrid, using the single-phase sensor definitions
12:02:46 INFO    Protocol version: 2.1
[...]

PR #534, opened 2025-10-16
Documentation updates for USR-W630 gateway

hamid-elaosta:
Add note to docs about using a 120ohm resistor on W630 gateway (#510)
Add configuration settings to docs for USR-W630 Modbus TCP - Modbus RTU setup

### DIFF
--- a/www/docs/guide/wiring.md
+++ b/www/docs/guide/wiring.md
@@ -60,7 +60,7 @@ Other tested adaptors
 
    > NOTE: The resistor legs should be trimmed to a more reasonable length to ensure they don't accidentally short together.
 
-   ![120ohm](../images/usr-w630-120ohm.jpg)
+   ![120ohm](../images/usr-w630-120ohm.jpg =200x200)
 
 
    #### USR-W630 Configuration (Wi-Fi)


www/docs/guide/wiring.md
```diff
@@ -56,6 +56,24 @@ Other tested adaptors
 
    Requires `READ_SENSORS_BATCH_SIZE` set to 8 or less
 
+   A 120 Ohm (brown-red-brown) resistor may be required for data communication between the adapter and inveter to occur without corruption. The resistor should be added between the A+B lines, such as in the example below.
+
+   > NOTE: The resistor legs should be trimmed to a more reasonable length to ensure they don't accidentally short together.
```
kellerza: You should trim the one in the picture 😆 

Can you add size info to the image tags? It seems `=400x400` is quite common on this page, but maybe the resistor should be `=200x200`?

kellerza (changes requested):
Thank you!

Interesting that you had to add it and in the fault-finding section there is a suggestion to remove it.

Can you measure the voltage drop over the resistor?

www/docs/guide/wiring.md
hamid-elaosta: Haha yeah, I've been meaning to trim it, it was temporary, but you know how it is; as soon as it was functional, I now don't want the downtime.

I'll try and find a moment to get a multimeter on it tomorrow and get the voltage drop.

200x200 seems right for this image, I'll add the size info (I tried it originally before submitting but it broke the rendering in my IDE so I removed it thinking perhaps it was for specific types).

kellerza: The size is not 100% standard markdown. The docs use a plugin to enable sizing

Thanks!

PR #556, opened 2025-10-26
Update getting-started.md

dvoytik:
Add note that installation of "Modbus TCP to Modbus RTU Gateway Add-on" is recommended.


kellerza:
It is only recommended when you connect directly to serial.

PR #585, opened 2026-01-15
Add connector system for shared Modbus connections & fix Sensor16 power spikesa

maslyankov:
## Summary
This PR introduces a connector system that enables multiple inverters to share a single Modbus connection, and fixes power value spikes in Sensor16 sensors.

**This branch has been in active use and testing for over a month, running perfectly with two SUN-30K-SG01HP3-EU-BM3.**

### 🔌 Shared Connector System
Allows multiple inverters to be monitored and controlled through a single RS485 bridge/hub. Previously, each inverter instance attempted to act as a separate Modbus master on the bus, causing communication conflicts. Now inverters can reference a shared connector:

```yaml
CONNECTORS:
  tcp_gateway:
    type: tcp
    host: 192.168.1.100
    port: 502

INVERTERS:
  - connector: tcp_gateway
    modbus_id: 1
    ha_prefix: inverter1
  - connector: tcp_gateway
    modbus_id: 2
    ha_prefix: inverter2
```

Each inverter maintains its own independent state while sharing the connection—works for both parallel inverter setups and separate inverters on the same bus.

### ⚡ Sensor16 Power Spike Fix
Fixed incorrect 16-bit/32-bit value interpretation that caused power sensors (Grid power, Load power, etc.) to report spikes or negative values:
- Values above ~32,767 no longer incorrectly become negative
- Eliminated spurious spikes from noise in the high register byte

## Backwards Compatibility
Existing configurations using the `PORT` option continue to work without changes.

### DIFF
--- a/www/docs/reference/multi-options.md
+++ b/www/docs/reference/multi-options.md
@@ -17,27 +17,27 @@ The `CONNECTORS` option allows you to define shared connections that can be used
 
 ```yaml
 CONNECTORS:
-  tcp_gateway:
-    type: tcp
-    host: 192.168.1.100
-    port: 502
-    driver: pymodbus
-    timeout: 10
+  - NAME: tcp_gateway
+    TYPE: tcp
+    HOST: 192.168.1.100
+    PORT: 502
+    DRIVER: pymodbus
+    TIMEOUT: 10
   
-  solarman_dongle:
-    type: solarman
-    host: 192.168.1.101
-    port: 8899
-    dongle_serial: 12345678
-    driver: solarman
-    timeout: 10
+  - NAME: solarman_dongle
+    TYPE: solarman
+    HOST: 192.168.1.101
+    PORT: 8899
+    DONGLE_SERIAL: 12345678
+    DRIVER: solarman
+    TIMEOUT: 10
   
-  serial_port:
-    type: serial
-    port: /dev/ttyUSB0
-    baudrate: 9600
-    driver: pymodbus
-    timeout: 10
+  - NAME: serial_port
+    TYPE: serial
+    PORT: /dev/ttyUSB0
+    BAUDRATE: 9600
+    DRIVER: pymodbus
+    TIMEOUT: 10
 ```
 
 ### Connector Types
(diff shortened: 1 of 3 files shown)


www/docs/reference/multi-options.md
```diff
@@ -11,6 +11,83 @@
...
+- `driver`: Driver to use (pymodbus, umodbus, solarman)
+- `timeout`: Connection timeout in seconds
+- `dongle_serial`: Required for solarman type
+- `baudrate`: Required for serial type
+
+### Example: Multiple Inverters with Shared Connector
+
+```yaml
+CONNECTORS:
+  tcp_gateway:
```
kellerza: I'm not sure a dict is supports validation in the Home Assistant addon schema (see `hass-addon-sunsynk-edge\config.yaml`).

If we change it to a list and use a `name` or `id` to identify the connector it should have the same effect.

It should also remove the need for the custom converter in options.py

maslyankov:
> Hello @maslyankov and @kellerza
> 
> With this new concept, are there any thoughts on how one can communicate with the inverters while there are batteries connected on the RS485 port. On some of the Sunsynk and Deye inverters the port used for comms to this integration is the same that is used for the battery so one has to choose.

As far as I remember, battery entities and data in general for batteries is fetched from the inverter itself. If there is support added for battery communication based on BMS protocols and not through the inverter, this approach with connectors should provide functionality for what you are asking.

Can you give more info on battery/bms model and inverter model?

maslyankov:
> > > Hello @maslyankov and @kellerza
> 
> > > With this new concept, are there any thoughts on how one can communicate with the inverters while there are batteries connected on the RS485 port. On some of the Sunsynk and Deye inverters the port used for comms to this integration is the same that is used for the battery so one has to choose.
> 
> > 
> 
> > As far as I remember, battery entities and data in general for batteries is fetched from the inverter itself. If there is support added for battery communication based on BMS protocols and not through the inverter, this approach with connectors should provide functionality for what you are asking.
> 
> > 
> 
> > Can you give more info on battery/bms model and inverter model?
> 
> 
> 
> Hello @maslyankov,
> 
> 
> 
> I am not worried about getting the data from the batteries. I am wondering if the system will function with having the batteries connected to the inverter on rs485 while still letting the integration connect to the inverter on the same RS485 bus.

I think it depends on how the inverter and battery communicate on the bus. One option is the inverter is master , battery is slave. This could be an issue for the sunsynk integration I guess. 
The changes in this repo would not affect this, so if it currently works for you, you can connect a second inverter and talk to both of them on the same bus, despite the battery being on it too. If it currently does not work, this PR will not help I think.

maslyankov:
Pls do not merge it yet. I would like to test it in my setups after the latest changes.

kellerza:
@maslyankov - I merged the lock PR #594

In the latest release, you could try share connectors by simply using the same `PORT` for multiple inverters (not documented yet)

Still open to accept the Sensor16 power spike PR, but we should probbaly make that opt-in using `SENSOR_OVERRIDES` and not a normal config option. With `SENSOR_OVERRIDES` you can set some attribute "new_firmware" or so and then run the new code instead of the old,- does it make sense?

PR #643, opened 2026-06-14
Update three_phase_hv.py

scenix26:
I Got this Inventer Deye SUN-80K-SG02HP3-EU-EM6  and got this Information from Deye from ==> MODBUS RTU三相储能通信规约V105.4-20240813最新 (002).doc

### DIFF
--- a/src/sunsynk/definitions/three_phase_hv.py
+++ b/src/sunsynk/definitions/three_phase_hv.py
@@ -50,18 +50,22 @@ SENSORS += (
 ##############
 SENSORS += (
     MathSensor(
-        (672, 673, 674, 675), "PV power", WATT, factors=(10, 10, 10, 10)
-    ),  # pv1,pv2,pv3,pv4 power
+        (672, 673, 674, 675,727, 728, 729, 730), "PV power", WATT, factors=(10, 10, 10, 10,10, 10, 10, 10)
+    ),  # pv1,pv2,pv3,pv4,pv5,pv6,pv7,pv8 power
 )
 
 ################
-# MPPT 1-4 Solar Power
+# MPPT 1-8 Solar Power
 ################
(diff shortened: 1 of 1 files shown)


kellerza:
Thanks @scenix26 !

Do you perhaps know if there is a register/way to know how many inverters are available in your setup?

scenix26:
Unfortunately there's no register that directly tells you how many inverters are in the system. What you can do instead is read Register 336 from each unit — Bit0 tells you if parallel is enabled, Bit1 whether it's master or slave, and Bits 10–15 give you the unit's Modbus SN in the chain. By scanning all addresses and checking those bits, you can reconstruct the full picture. If your setup has a monitoring gateway/dongle, that might expose an aggregated device count — but that would be a separate protocol, not the standard Modbus RTU one. @kellerza

kellerza:
@copilot Please fix the formatting issue ruff identified

src/sunsynk/definitions/three_phase_hv.py
```diff
@@ -50,18 +50,22 @@
 ##############
 SENSORS += (
     MathSensor(
-        (672, 673, 674, 675), "PV power", WATT, factors=(10, 10, 10, 10)
-    ),  # pv1,pv2,pv3,pv4 power
+        (672, 673, 674, 675,727, 728, 729, 730), "PV power", WATT, factors=(10, 10, 10, 10,10, 10, 10, 10)
```
kellerza:
```suggestion
        (672, 673, 674, 675, 727, 728, 729, 730),
        "PV power",
        WATT,
        factors=(10, 10, 10, 10, 10, 10, 10, 10),
    ),  # pv1,pv2,pv3,pv4,pv5,pv6,pv7,pv8 power
```

PR #661, opened 2026-07-26
Support for SUN-10K-SG05LP3-EU-SM2

Cosik:
Hi,

I'm fresh owner of SUN-10K-SG05LP3-EU-SM2. Yesterday I connect RS485 and start playing with it. I noticed in log that line

`[05:35:13] WARNING device_type: Unknown register value 1280. Consider extending the definition with a PR. https://github.com/kellerza/sunsynk/tree/main/src/sunsynk/definitions`

So here is my PR.

Last thing, I've name this PR as support for this device. One thing which left is fix for SOC Battery temperature. It seems that it is working without any offset. I don't know full history of add-on so I'm open to how to do it in correct way. So far I've done it by overwritten sensor by custom sensor, but I think that this could be useful for other users.


kellerza:
Thanks!

You can also try SENSOR_OVERRIDES if this is only a couple of sensors where parameters need to change. In many cases easier to use than a complete custom sensor

PR #669, opened 2026-08-12
Offer time values across midnight

### LINKED ISSUE #667: TimeRWSensor.available_values never offers times past midnight, so a program slot can never move across midnight
Describe the issue/bug
Expected behavior
Every 15-minute slot in the window 23:00 → 5:00, i.e.
23:00, 23:15, 23:30, 23:45, 0:00, 0:15, … , 4:45, 5:00. The candidate list
should span the same range as the (possibly wrapped) bounds, i.e. 48 hours
instead of 24.

Suggested fix — three small changes in available_values:

day = list(range(0, 48 * 60, step_minutes)) — the actual fix.
Roll the current value along too (if val < minv: val += 24 * 60), otherwise
a current value that already sits past midnight is appended below minv and
shows up twice.
Deduplicate the rendered strings — SSTime.str_value folds back onto a 24h
clock (hours % 24), so 0:00 and 24:00 render identically. This also
fixes an existing duplicate 0:00 in the minv == maxv (full-day) case.
    def available_values(self, step_minutes: int, state: InverterState) -> list[str]:
        """Get the available values for this sensor."""
        # Two days of candidates, since maxv can wrap past midnight below
        day = list(range(0, 48 * 60, step_minutes))
        minv = SSTime(strv=str(state.get(self.min, "0:00"))).minutes if self.min else 0
        maxv = SSTime(strv=str(state.get(self.max, "0:00"))).minutes if self.max else 0
        if minv >= maxv:
            maxv += 24 * 60
        opt = [minv, *[v for v in day if (v > minv and v < maxv)], maxv]
        val = SSTime(strv=str(state.get(self, "0:00"))).minutes
        if val < minv:  # the current value already wrapped past midnight
            val += 24 * 60
        if val not in opt:
            opt.append(val)
        opt.sort()
        # str_value wraps to a 24h clock, so dedupe (e.g. minv 0:00 & maxv 24:00)
        return list(dict.fromkeys(SSTime(minutes=m).str_value for m in opt))
With that patch the example above returns the full 23:00 → 5:00 range, and the
existing test suite (src/tests/sunsynk) still passes unchanged — including the
existing min 23:00 / max 24:00 case in test_rwsensors.py, which is not a
wrap (minv < maxv) and keeps its current output.

Suggested regression test:

def test_time_rw_wrap_over_midnight() -> None:
    """A time sensor whose window wraps past midnight offers the wrapped slots."""
    state = InverterState()
    s = TimeRWSensor(60, "prog1", factor=0.1)
    s.min = TimeRWSensor(50, "prog6", factor=0.1)
    s.max = TimeRWSensor(70, "prog2", factor=0.1)
    state.track(s, s.min, s.max)
    state.update({50: 2300, 60: 100, 70: 500})  # 23:00 .. 1:00 .. 5:00

    values = s.available_values(60, state)
    assert values == ["23:00", "0:00", "1:00", "2:00", "3:00", "4:00", "5:00"]
    assert len(values) == len(set(values))
It fails on main (returns ['1:00', '23:00', '5:00']) and passes with the fix.

Logs (if applicable)
n/a — reproducible from code alone, see the snippet above.
Environment: sunsynk main @ 43d4975 (pyproject.toml version 0.9.3); the
same code is in the released add-on (0.10.1). Deye SUN-12K-SG05LP3-EU
(three-phase LV), SENSOR_DEFINITIONS: three-phase, PROG_TIME_INTERVAL: 15.

Happy to open a PR with the patch and test, and to verify on real hardware.

luantaraschi:
## Summary

- Extend time candidates to 48 hours when a configured window crosses midnight.
- Keep the current value in the wrapped interval and remove duplicate rendered times.
- Add a regression test for a 23:00 to 05:00 window.

Closes #667

## Testing

- `pytest src/tests/sunsynk` (71 passed)
- `pytest src/tests/sunsynk/test_rwsensors.py` (10 passed)
- `ruff check src/sunsynk/rwsensors.py src/tests/sunsynk/test_rwsensors.py`
- `ruff format --check src/sunsynk/rwsensors.py src/tests/sunsynk/test_rwsensors.py`

### DIFF
--- a/src/sunsynk/rwsensors.py
+++ b/src/sunsynk/rwsensors.py
@@ -231,17 +231,20 @@ class TimeRWSensor(RWSensor):
 
     def available_values(self, step_minutes: int, state: InverterState) -> list[str]:
         """Get the available values for this sensor."""
-        day = list(range(0, 24 * 60, step_minutes))
+        day = list(range(0, 48 * 60, step_minutes))
         minv = SSTime(strv=str(state.get(self.min, "0:00"))).minutes if self.min else 0
         maxv = SSTime(strv=str(state.get(self.max, "0:00"))).minutes if self.max else 0
-        if minv >= maxv:
+        wraps_midnight = minv >= maxv
+        if wraps_midnight:
             maxv += 24 * 60
         opt = [minv, *[v for v in day if (v > minv and v < maxv)], maxv]
         val = SSTime(strv=str(state.get(self, "0:00"))).minutes
+        if wraps_midnight and val < minv:
+            val += 24 * 60
         if val not in opt:
             opt.append(val)
         opt.sort()
-        return [SSTime(minutes=m).str_value for m in opt]
+        return list(dict.fromkeys(SSTime(minutes=m).str_value for m in opt))
 
     @property
     def dependencies(self) -> list[Sensor]:
--- a/src/tests/sunsynk/test_rwsensors.py
+++ b/src/tests/sunsynk/test_rwsensors.py
@@ -249,6 +249,21 @@ def test_time_rw(state: InverterState) -> None:
     assert s.available_values(15, state) == ["2:03", "2:10", "2:15", "2:22"]
 
 
+def test_time_rw_wrap_over_midnight() -> None:
+    """A time sensor whose window wraps past midnight offers the wrapped slots."""
+    state = InverterState()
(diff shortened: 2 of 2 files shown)


kellerza:
Thank you @luantaraschi

PR #687, opened 2026-08-31
Apply READ_ATTEMPTS to the identity read

### LINKED ISSUE #686: Identity read (regs 0-7) ignores READ_ATTEMPTS, so one dropped reply aborts the add-on
## Issue related to

Sunsynk Home Assistant Add-On

### Describe the issue/bug

`AInverter.read_identity()` builds an `Identity` Component over `self.inv.unit` and calls
`async_update()`. That reads holding registers 0-7 directly off the unit, so it never goes through
`Sunsynk.read_holding_registers()` — the only place `READ_ATTEMPTS` is applied. There is no retry
anywhere in `modbus_connection.model`, so the identity read gets exactly one attempt no matter what
`READ_ATTEMPTS` is set to.

Because `main_loop()` turns any `ConnectionError` out of `ist.connect()` into `return 2`, one
dropped reply on one inverter aborts the whole add-on.

The asymmetry is visible in the log — every other read failure carries `[attempt n/3]`, the fatal
one does not:

```conf
[11:33:14] ERROR   Read register 552 x7: ModbusTimeoutError: read_holding_registers(552, 7): Response timeout after 2 seconds [attempt 1/3]
[11:33:19] INFO    No response on the Modbus interface /dev/ttyACM1: read_holding_registers(0, 8): Response timeout after 2 seconds
[11:33:19] CRITICAL This Add-On will terminate in 30 seconds, use the Supervisor Watchdog to restart automatically.
```

With three inverters this becomes a restart loop: whichever port was left mid-transaction by the
previous SIGKILL is the one that misses next, and the ~6 s watchdog restart never gives the port
time to settle. Ports rotate between runs — 11:33:02 `/dev/ttyACM2`, 11:33:19 `/dev/ttyACM1`.

Lowering `TIMEOUT` makes it much more likely, since the single-shot read has no retries to absorb a
miss, but it also happens at `TIMEOUT: 3`.

**Expected behavior** The identity read honours `READ_ATTEMPTS` like every other holding-register
read, including the serial flush between attempts.

### Your environment

- Addon:
  - Name: sunsynk-edge
  - Version: 246721b
- Inverter:
  - Make: Sunsynk
  - Model: high-voltage three-phase hybrid, 3 x 50 kW (identity reports device type `0x6`,
    protocol 1.4)
- Adaptor details:
  - RS485 over USB CDC, `/dev/ttyACM0` .. `/dev/ttyACM2`, one adaptor per inverter

### Your configuration

```yaml
INVERTERS:
  - SERIAL_NR: "**********"
    HA_PREFIX: SS1
    MODBUS_ID: 1
    PORT: /dev/ttyACM0
  - SERIAL_NR: "**********"
    HA_PREFIX: SS2
    MODBUS_ID: 2
    PORT: /dev/ttyACM1
  - SERIAL_NR: "**********"
    HA_PREFIX: SS3
    MODBUS_ID: 3
    PORT: /dev/ttyACM2
SENSOR_DEFINITIONS: three-phase-hv
READ_ALLOW_GAP: 2
READ_SENSORS_BATCH_SIZE: 20
READ_MESSAGE_SPACING: 0.05
READ_ATTEMPTS: 3
TIMEOUT: 2
```

### Logs

```conf
[11:33:10] INFO    sunsynk library - commit 246721b - last stable version 1.2.0
[11:33:10] INFO    Connecting to /dev/ttyACM0
[11:33:11] INFO    Identity: device_type=High-voltage three-phase hybrid (0x6) protocol=1.4 serial=****18594
[11:33:14] ERROR   Read register 552 x7: ModbusTimeoutError: read_holding_registers(552, 7): Response timeout after 2 seconds [attempt 1/3]
[11:33:17] INFO    Connecting to /dev/ttyACM1
[11:33:17] INFO    Reading inverter identity (registers 0-7)
[11:33:19] INFO    No response on the Modbus interface /dev/ttyACM1: read_holding_registers(0, 8): Response timeout after 2 seconds
[11:33:19] CRITICAL This Add-On will terminate in 30 seconds, use the Supervisor Watchdog to restart automatically.
[09:33:19] WARNING: Addon resource not ready 2
s6-supervise sunsynk: warning: finish script lifetime reached maximum value - sending it a SIGKILL
[11:33:25] INFO: Starting the Sunsynk add-on
```

heinrich321:
Fixes #686

`AInverter.read_identity()` built the `Identity` Component over `self.inv.unit`, so the registers
0-7 read went straight to the unit and never through `Sunsynk.read_holding_registers()` — the only
place `READ_ATTEMPTS` is applied. Nothing in `modbus_connection.model` retries, so the identity read
got a single attempt whatever `READ_ATTEMPTS` said, and `main_loop()` turns the resulting
`ConnectionError` into `return 2`.

Passing the `Sunsynk` rather than its unit reuses the path that already exists: same
`read_holding_registers(address, count)` signature, plus the retries, the serial flush between
attempts, and the `timeouts` counter behind the RS485 timeout entity. A failed identity read now
also logs `[attempt n/m]` like every other read, which is what made this hard to spot in user logs.

The second commit is needed for the first to report correctly. The `ModbusProtocolError` branch in
`read_holding_registers` logged but never appended to `errs`, so a group where every attempt
desynced raised `ValueError: second argument (exceptions) must be a non-empty sequence` instead of
the Modbus failure. That is reachable today on a noisy RS485 link, and the identity read now shares
the path.

### Scope

Only `Identity` moves onto this path, and it reads holding registers only — the one space `Sunsynk`
exposes. Flagged with a `ponytail:` comment naming that ceiling.

### Verification

Each test fails with its fix reverted and passes with it:

- `test_read_identity_retries_dropped_reply` — `ModbusTimeoutError` on the first attempt,
  registers on the second; asserts two calls and a decoded serial.
- `test_ss_protocol_error_raises_exception_group` — every attempt a `ModbusProtocolError`; asserts
  the group carries them.

`ruff check`, `ruff format --check`, `zuban check src` and `pytest` are clean (128 passed).


src/ha_addon_sunsynk_multi/a_inverter.py
```diff
@@ -153,8 +153,11 @@ async def lifecycle_attempt_recovery(self) -> None:
 
     async def read_identity(self) -> Identity:
         """Read device type, protocol, and serial via the Identity Component."""
-        # HoldingUnit duck-types ModbusUnit for FC03; SolarmanUnit is the same shape.
-        identity = Identity(self.inv.unit)  # type: ignore[arg-type]
+        # Read through Sunsynk, not self.inv.unit: same read_holding_registers
+        # signature, but with READ_ATTEMPTS retries and the serial flush.
+        # ponytail: only holds while Identity reads holding registers only. A field
+        # in another space would need the unit's read_coils/read_input_registers.
+        identity = Identity(self.inv)  # type: ignore[arg-type]
```
kellerza, line 160: Where do you get the modbus ID from? This is part of the unit?

heinrich321: Yes, the modbus ID is on the unit, and it stays there.

`Sunsynk.read_holding_registers` just calls `self.unit.read_holding_registers` (sunsynk.py:170), so passing `self.inv` instead of `self.inv.unit` still lands on the same unit that driver.py:82 builds with `conn.for_unit(iopt.modbus_id)`. Same for solarman, where the id sits on `SolarmanUnit(server_id=iopt.modbus_id)` at driver.py:67.

The only thing added in between is the retry, the serial flush and the timeouts counter.

I checked it with a real connection and `for_unit(3)`, recording what actually hits the wire for the old and the new call:

```
wire calls: [{'unit_id': 3, 'address': 0, 'count': 8},
             {'unit_id': 3, 'address': 0, 'count': 8}]
serial before: 2303218594   after: 2303218594
```

First one is `Identity(unit)`, second is `Identity(ss)`. Same unit id, same result.

Happy to add an assert on the unit id to the test if you want that locked in.

kellerza: Ok, but we still pass something that is not really a HoldingUnit

Can we rather add this to `sunsynk.py`?

Add a tiny retrying unit facade on Sunsynk, and pass that to Identity:

```python
@dataclass(frozen=True, slots=True)
class _RetryingUnit:
    """HoldingUnit that routes FC03 through Sunsynk retry policy."""
    ss: Sunsynk
    @property
    def connected(self) -> bool:
        return self.ss.unit.connected
    async def read_holding_registers(self, address: int, count: int) -> list[int]:
        return list(await self.ss.read_holding_registers(address, count))
    async def write_registers(self, address: int, values: list[int]) -> None:
        await self.ss.unit.write_registers(address, values)
```

Then in read_identity:
```python
identity = Identity(self.inv.retrying_unit)  # property returning _RetryingUnit(self)
await identity.async_update()
```

That gives you:

- Correct modbus id (still ends up on self.inv.unit)
- Sunsynk retry/flush behavior
- A object that actually matches HoldingUnit
- No coupling of Identity to the whole Sunsynk API

